PR #7, opened 2018-02-01
Draft lesson zero, hello world version.

adziu-tieto:
Lesson zero again, without information overload.


course/lesson_00_the_setup/README.md
```diff
@@ -2,10 +2,26 @@
 #### introduction
 - Install [Python 3.6](https://www.python.org/downloads/release/python-364/),
 - Install [PyCharm Community Edition IDE](https://www.jetbrains.com/pycharm/download/) - strongly recommended,
+- `pip install bpython ipython pytest ipdb`,
```
jedzej: bpython crashes on windows with error:  No module named '_curses'
Instructions on the internet suggest to download and install it separately to solve this dependency. My proposal is to remove bpython it and leave ipython only.

PR #8, opened 2018-02-02
Git videos and expected problems (FAQ) section.

adziu-tieto:
We spoke about some possible issues, that the users will have. Maybe it's a good idea to address those. If not immediately, then at least as they show up. If somebody really asks about vim, then we can provide the link.


kuszcjan:
Let's merge this

PR #26, opened 2018-02-15
lesson 0 instruction update

jedzej:
Aligning branching convention with directories naming convention


adziu-tieto:
I proposed to my group that they use my login there, I get their name already by their directory. It's kind of late to change that. I'll commit my proposal in this branch.

jedzej:
Adrian's idea is fine for me. I've fixed one omitted line to keep the consistency.

PR #66, opened 2018-02-25
lesson_01 problems solved

akpoik:
@gradopaw


akpoik:
@gradopaw

akpoik:
@gradopaw

students/kulon_aleksander/lesson_01/Problem «First digit after decimal point» (Easy).py
```diff
@@ -1,2 +1,2 @@
 f = float(input())
-print(((f % 1 - f % 0.1) * 10)//1)
\ No newline at end of file
+print(((f % 1 - f % 0.1) * 10) // 1)
```
gradopaw: Did you check this? As I can see it is not working properly :)

akpoik:
@gradopaw 

problem_first_digit_after_decimal_point.py is stil giving errors(but values match) on sneakify, I do not know why.

gradopaw:
That's because of the usual floating point rounding issues. Everything is going to be wrong after the first modulo operation (f % 1),  where for the input value 7.9 you getting in a result 0.9000000000000004 so it differs from the result you expect - it's not your fault. The whole algorithm to calculate this task is proper. Your code passed all snakify tests because there is no such test for similar input value for which I have performed my own test. One of the ways to resolve this issue is:
f = float(input())
print(int((f- int(f)) * 10))

I think we can proceed with the merge.

PR #78, opened 2018-02-26
Lesson 01 basics

hatan:
@kamwar here is my lession 1.


students/kosarzewski_maciej/lesson_01_basics/digital_clock.py
```diff
@@ -0,0 +1,4 @@
+# Read an integer:
+a = int(input())
+# Print a value:
+print(str(a//60) + " " + str(a%60))
```
kamwar: PEP 8: missing whitespace around modulo operator

PEP8 Warning in PyCharm is underlined with the orange curvy line. If you hover your mouse pointer over this curvy line, the detailed information about the error appears at the tooltip.
image


Even better soultion, always check you code before PR is created with inspect tool. Just select 'Code' -> 'Inspect Code'  in PyCharm IDE
image

students/kosarzewski_maciej/lesson_01_basics/apple_sharing.py
```diff
@@ -0,0 +1,6 @@
+# Read the numbers like this:
+students = int(input())
```
kamwar: It will be great to include text description whenever user is prompt for input
students = int(input("Provide numer of students: "))

hatan:
@kamwar please decline this PR.

PR #82, opened 2018-02-27
Tasks from Lesson_01

piotrbiegon:


kuszcjan:
There are different naming conventions for python files:

"So PEP 0008 tells you that:
modules (filenames) should have short, all-lowercase names, and they can contain underscores;"
e.g.: appleShare.py -> apple_share.py
https://softwareengineering.stackexchange.com/questions/308972/python-file-naming-convention

students/biegon_piotr/lesson_01_basics/Sum_of_digits.py
```diff
@@ -0,0 +1,5 @@
+print("Sum of digits\n")
+
+number = int(input("Enter a three-digit number: "))
+
+print("\nSum of a three-digit number is:", (sum([int(digit) for digit in str(number)])))
```
kuszcjan: unnecessary parenthesis.

It's okay, but in the real world, we should avoid casting to string and then back to int.

students/biegon_piotr/lesson_01_basics/Apple_sharing.py
```diff
@@ -0,0 +1,10 @@
+print("Apple sharing\n")
+
+N = int(input("Enter the number of students: "))
+K = int(input("Enter the number of apples: "))
+
+a = K // N
+b = K % N
+
+print("\nEach student will get %d apples" %a)
```
kuszcjan: It's okay, but please check out new style formatting:
https://pyformat.info/

kuszcjan:
Good.
Just take into account my comments for next PR.

PR #83, opened 2018-02-27
Added lesson_01_basics practice projects

swietmon:
@kamwar Please check this

### DIFF
--- a/students/swietczak_monika/lesson_01_basics/apple_sharing.py
+++ b/students/swietczak_monika/lesson_01_basics/apple_sharing.py
@@ -1,7 +1,7 @@
 # Read the numbers like this:
-n = int(input())
-k = int(input())
+n = int(input("How many students eat apples? "))
+k = int(input("How many apples? "))
 # Print the result with print()
 
-print(k//n)
-print(k%n)
+print(k // n)
+print(k % n)
--- a/students/swietczak_monika/lesson_01_basics/apple_sharing.py
+++ b/students/swietczak_monika/lesson_01_basics/apple_sharing.py
@@ -1,0 +1,7 @@
+# Read the numbers like this:
+n = int(input())
+k = int(input())
+# Print the result with print()
+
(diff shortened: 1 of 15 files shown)


students/swietczak_monika/lesson_01_basics/apple_sharing.py
```diff
@@ -0,0 +1,7 @@
+# Read the numbers like this:
+n = int(input())
+k = int(input())
+# Print the result with print()
+
+print(k//n)
+print(k%n)
```
kamwar: Please remove all PEP8 warning in your PR.

Warning is underlined with the orange curvy line. If you hover your mouse pointer over this curvy line, the detailed information about the error appears at the tooltip.
image


Even better soultion, always check you code before PR is created with inspect tool. Just select 'Code' -> 'Inspect Code'  in PyCharm IDE
image

students/swietczak_monika/lesson_01_basics/area_of_right_angled_triangle.py
```diff
@@ -0,0 +1,6 @@
+# Read the numbers b and h like this:
+b = int(input())
+h=int(input())
```
kamwar: PEP8

students/swietczak_monika/lesson_01_basics/car_route.py
```diff
@@ -0,0 +1,7 @@
+# Read an integer:
+import math
+N = int(input())
```
kamwar: Capital letters should be used only for constants e.g.
MY_NAME = "Kamil"

Please use lower case instead:
n = int(input())

kamwar: It will be great to include text description whenever user is prompt for input
n = int(input("How many kilometers per day: "))

swietmon:
@kamwar changes done

PR #101, opened 2018-03-01
lesson_01 tasks

jemiekam:
@kamwar


students/jemielity_kamil/lesson_01_basics/apple_sharing.py
```diff
@@ -0,0 +1,10 @@
+import math
+
+N = int(input("Number of students: "))
```
kamwar: Capital letters should be used only for constants e.g.
MY_NAME = "Kamil"

Please use lower case instead:
n = int(input("Number of students: "))

students/jemielity_kamil/lesson_01_basics/apple_sharing.py
```diff
@@ -0,0 +1,10 @@
...
+K = int(input("Amount of apples: "))
+
+amountOfApplesPerStudent = math.floor(K/N)
```
kamwar: Please use underscores instead of camel case for variable name in python
apples_per_student = math.floor(k / n)

students/jemielity_kamil/lesson_01_basics/car_route.py
```diff
@@ -0,0 +1,6 @@
+import math
+N = int(input("How many kilometers per day: "))
+M = int(input("How many kilometres: "))
+
+days = M / N
+print(math.ceil(days))
```
kamwar: According to PEP8 file should always end with a new line

jemiekam:
@kamwar - added new commit with fixes

students/jemielity_kamil/lesson_01_basics/tens_digit.py
```diff
@@ -0,0 +1,5 @@
+
+integer = int(input("Write an integer: "))
+
+print(str(integer)[-2])
```
kamwar: I believe this solution wasn't accepted by Snakify. Please add a comment if you can't find the correct solution.

Try to use 'integer // 2' in your code

students/jemielity_kamil/lesson_01_basics/total_cost.py
```diff
@@ -0,0 +1,9 @@
+import math
+dollars = int(input('How many dollars: '))
+cents = int(input('How many cents: '))
+number_of_cupcakes = int(input('Number of cupcake: '))
+
+wallet = str(dollars) + '.' + str(cents)
+cost_of_cupcakes = float(wallet) * number_of_cupcakes
+print(str(cost_of_cupcakes).split('.')[0] + " " + str(cost_of_cupcakes).split('.')[1])
+
```
kamwar: Similar like in the above comment, please ensure that you can pass all Snakify tests

students/jemielity_kamil/lesson_01_basics/total_cost.py
```diff
@@ -0,0 +1,9 @@
+import math
+dollars = int(input('How many dollars: '))
+cents = int(input('How many cents: '))
+number_of_cupcakes = int(input('Number of cupcake: '))
+
+wallet = str(dollars) + '.' + str(cents)
```
kamwar: Check what happens if cents = 6.
It might be better solution to count total price in cents instead of using float value

jemiekam:
@kamwar - next fixes

PR #106, opened 2018-03-01
Lession 01: BASIC

hatan:
@kamwar lesson 01, second attempt  :)


students/kosarzewski_maciej/lesson_01_basics/fractional_part.py
```diff
@@ -0,0 +1,4 @@
+# Read a float:
+a = float(input())
+# Print a value
+print (a - int(a))
```
kamwar: PEP8 warning

Please check PEP8 warnings with 'Code' -> 'Inspect code'

students/kosarzewski_maciej/lesson_01_basics/total_cost.py
```diff
@@ -0,0 +1,7 @@
+# Read an integer:
+A = int(input())
```
kamwar: Capital letters should be used only for constants e.g.
MY_NAME = "Kamil"

Please use lower case with underscores for variable names instead:

students/kosarzewski_maciej/lesson_00_the_setup/importable_module.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
```
kamwar: Lesson 00 should not be a part of this PR. Lesson 00 is already delivered into jedzej/tietopythontraining-basic

students/kosarzewski_maciej/lesson_01_basics/fractional_part.py
hatan: For my PyCharm is ok :\

kamwar: There should be no space after print statement. For some reason your pyCharm doesn't check PEP8. This time it's ok but try to follow PEP8 style guide next time.

PR #108, opened 2018-03-01
Add solutions for lesson_02_flow_control

mielcszy:


mielcszy:
@kuszcjan - done, please review

students/mielczarek_szymon/lesson_02_flow_control/lost_card.py
```diff
@@ -0,0 +1,10 @@
+n = int(input())
+sum_rem = 0
+sum_all = 1
+
+for i in range(2, n + 1):
+    a = int(input())
+    sum_rem += a
```
kuszcjan: could be simplified to:
` sum_rem += int(input())`

students/mielczarek_szymon/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,8 @@
+n = int(input())
+cnt = 0
+
+for i in range(n):
```
kuszcjan: "i" variable is not used in this context.
by convetion this should be changed to underscore '_'

"As a general purpose "throwaway" variable name to indicate that part of a function result is being deliberately ignored"
from:
https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python

students/mielczarek_szymon/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,8 @@
+n = int(input())
+cnt = 0
```
kuszcjan: it also would be nice to use more elaborate names, like "num_zeroes "

kuszcjan:
some things could be done better, but it's good enough ;)

PR #110, opened 2018-03-02
lesson02+lesson03 tasks

jemiekam:
@kamwar


jemiekam:
@kamwar added lesson03 tasks

students/jemielity_kamil/lesson_02_flow_control/chocolate_bar.py
```diff
@@ -0,0 +1,19 @@
+
+n = int(input("Length of chocolate: "))
+m = int(input("Width of chocolate "))
+k = int(input("Number of squares: "))
+
+if k > n*m:
```
kamwar: It's not stated in PEP8 but it is good practice to use space between math operators 'if k > n * m:'

students/jemielity_kamil/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,19 @@
...
+
+ending_column = int(input('Ending column: '))
+ending_row = int(input('Ending row: '))
+
+if abs(starting_column - ending_column) == 2:
+    if ending_row == starting_row + 1 or starting_row - 1:
+        print('YES')
+    else:
+        print('NO')
+elif abs(starting_row - ending_row) == 2:
```
kamwar: My first impression is that 'elif abs(starting_row - ending_row) == 2:' should be merged with 'if abs(starting_column - ending_column) == 2:'

I noticed that this code doesn't pass snakify test. Are you using Snakify tests to verify your solution? If not, please do

students/jemielity_kamil/lesson_02_flow_control/ladder.py
```diff
@@ -0,0 +1,8 @@
+
+number = int(input('Integer: '))
+
+for row in range(1, number+1):
+    for i in range(1, row+1):
+        print(i, end='')
+
+    print(sep='')
```
kamwar: Isn't this print just for an empty line? If yes then it enough to print empty string print('')

students/jemielity_kamil/lesson_03_functions/README.md
```diff
@@ -0,0 +1,10 @@
+### Lesson 3 - Functions
```
kamwar: Please always create a new PR for a new lesson

PR #112, opened 2018-03-02
lesson 02 flow control

AxelNowaczyk:
@tomekDiakowTieto


students/axel_nowaczyk/lesson_02_flow_control/Factorial.py
```diff
@@ -0,0 +1,10 @@
+def factorial():
+    number = int(input())
+    product = 1
+    for i in range(2, number+1):
```
tomekDiakowTieto: Good

students/axel_nowaczyk/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,20 @@
+def number_of_elements_equal_to_the_maximum():
+    inputs = []
+    while True:
```
tomekDiakowTieto: Try to use other condition (without True and break)

students/axel_nowaczyk/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,20 @@
...
+        number = int(input())
+        if number == 0:
+            break
+        inputs.append(number)
+    sorted_inputs = sorted(inputs, reverse=True)
+    max = sorted_inputs[0]
```
tomekDiakowTieto: max is built in function. Be carefull:
>>> max
<built-in function max>
>>> max([1,5,2])
5
>>> max = 20
>>> max([1,5,2])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'int' object is not callable
>>>

students/axel_nowaczyk/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
AxelNowaczyk: Similar condition I have found on official wiki.python.org website as common practice. Link:  https://wiki.python.org/moin/WhileLoop

PR #116, opened 2018-03-04
lesson 01

PluciennikEdyta:
@adziu-tieto please check


course/lesson_01_basics/fractional_part.py
```diff
@@ -0,0 +1,4 @@
+# Read an integer:
+a = float(input())
+# Print a value:
+print(a % 1)
```
adziu-tieto: Cool, I wouldn't expect that to work.

PR #131, opened 2018-03-05
Lesson_02 solutions

synulewar:
@kuszcjan

### DIFF
--- a/students/synowiec_krzysztof/lesson_02_flow_control/chocolate_bar.py
+++ b/students/synowiec_krzysztof/lesson_02_flow_control/chocolate_bar.py
@@ -1,0 +1,13 @@
+def main():
+    m = int(input())
+    n = int(input())
+    k = int(input())
+    max = m*n
+
+    if (k % m == 0  or k % n == 0) and k < max:
+        print("YES")
+    else:
+        print("NO")
+
+if __name__ == '__main__':
+    main()
--- a/students/synowiec_krzysztof/lesson_02_flow_control/lost_card.py
+++ b/students/synowiec_krzysztof/lesson_02_flow_control/lost_card.py
@@ -1,0 +1,9 @@
+def main():
+    numberOfCards = int(input())
+    sum = numberOfCards * (numberOfCards + 1) // 2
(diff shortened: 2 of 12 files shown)


students/synowiec_krzysztof/lesson_02_flow_control/chocolate_bar.py
```diff
@@ -0,0 +1,13 @@
+def main():
+    m = int(input())
+    n = int(input())
+    k = int(input())
+    max = m*n
```
kuszcjan: variable name `max` shadows python built-in function `max()`
You can check those at https://docs.python.org/3/library/functions.html

students/synowiec_krzysztof/lesson_02_flow_control/lost_card.py
```diff
@@ -0,0 +1,9 @@
+def main():
+    numberOfCards = int(input())
+    sum = numberOfCards * (numberOfCards + 1) // 2
```
kuszcjan: Variable name `sum` shadows python built-in function `sum()`
https://docs.python.org/3/library/functions.html
Please avoid that in future.

students/synowiec_krzysztof/lesson_02_flow_control/queen_move.py
```diff
@@ -0,0 +1,20 @@
+def main():
+ startRow = int(input())
```
kuszcjan: Looks like there is only one space used for main() function.
Spacing should be the same for all files within a project. Standard practice is to use is 4 spaces.

students/synowiec_krzysztof/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,17 @@
...
+            max = n
+        elif n > secondMax:
+            secondMax = n
+
+        n = int(input())
+    print(secondMax)
+
+
+if __name__ == '__main__':
+    main()
```
kuszcjan: Lack of empty newline at the end of file.

PR #146, opened 2018-03-06
Lesson 1/2: Added python files

tomekktomekk:
@adziu Hi, please check my answers :)


students/kondel_tomasz/lesson_01_basics/Clock_face_1.py
```diff
@@ -0,0 +1,21 @@
...
+    if int(ang) == ang:
+        print(int(ang))
+    else:
+        print(ang)
+elif math.fabs(per) < 1:
+    print(round(ang, 3))
+elif math.fabs(per) < 100:
+    print(round(ang, 5))
+else:
+    print(round(ang, 8))
```
adziu-tieto: The thing you are looking for is `print('{:g}'.format(ang))`. I don't think they expected precision to be handled in that task. Quite a feat to make it pass, anyway :+1:

students/kondel_tomasz/lesson_01_basics/Car_route.py
```diff
@@ -0,0 +1,4 @@
+from math import ceil
+N = int(input())
+M = int(input())
+print(ceil(M / N))
```
adziu-tieto: It's a good practice to import `math` and use qualified names.

students/kondel_tomasz/lesson_01_basics/First_digit_after_decimal_point.py
```diff
@@ -0,0 +1,7 @@
+a = input()
+b = a.split('.')
+if len(b) > 1:
+    c = b[1]
+    print(c[0])
+else:
+    print(0)
```
adziu-tieto: Works, but it's easier with just `int(a * 10 % 10)`

students/kondel_tomasz/lesson_01_basics/Fractional_part.py
```diff
@@ -0,0 +1,7 @@
+import re
+a = input()
+b = a.split('.')
+if len(b) > 1:
+    print('0.' + str(b[1]))
+else:
+    print(0)
```
adziu-tieto: Again, and also to my surprise `a % 1` works too or using `math.fmod`.

PR #148, opened 2018-03-06
Tasks from Lesson_02

piotrbiegon:
@kuszcjan

### DIFF
--- a/students/biegon_piotr/lesson_02_flow_control/knight_move.py
+++ b/students/biegon_piotr/lesson_02_flow_control/knight_move.py
@@ -1,0 +1,22 @@
+print("Knight move\n")
+
+print("Enter the start coordinates:")
+x1 = int(input("x1 = "))
+y1 = int(input("y1 = "))
+
+print("\nEnter the final coordinates:")
+x2 = int(input("x2 = "))
+y2 = int(input("y2 = "))
+
+print("\nIs it possible to move the knight in one move to the coordinates given above?")
+
+if x2 == x1 - 2 and (y2 == y1 + 1 or y2 == y1 - 1):
+    print("YES")
+elif x2 == x1 - 1 and (y2 == y1 + 2 or y2 == y1 - 2):
+    print("YES")
+elif x2 == x1 + 1 and (y2 == y1 + 2 or y2 == y1 - 2):
+    print("YES")
+elif x2 == x1 + 2 and (y2 == y1 + 1 or y2 == y1 - 1):
+    print("YES")
+else:
+    print("NO")
--- a/students/biegon_piotr/lesson_02_flow_control/lost_card.py
+++ b/students/biegon_piotr/lesson_02_flow_control/lost_card.py
@@ -1,0 +1,17 @@
+print("Lost card\n")
+
+N = int(input("Enter the number of cards: "))
+
+x = 0
+y = 0
+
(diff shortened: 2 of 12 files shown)


students/biegon_piotr/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,22 @@
+print("Knight move\n")
+
+print("Enter the start coordinates:")
+x1 = int(input("x1 = "))
```
kuszcjan: it's also possible to take input in pairs, like:
`x1, y1 = int(input()), int(input())`

students/biegon_piotr/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,22 @@
...
+y1 = int(input("y1 = "))
+
+print("\nEnter the final coordinates:")
+x2 = int(input("x2 = "))
+y2 = int(input("y2 = "))
+
+print("\nIs it possible to move the knight in one move to the coordinates given above?")
+
+if x2 == x1 - 2 and (y2 == y1 + 1 or y2 == y1 - 1):
```
kuszcjan: The more elegant and readable solution would be:
```
if (x2 == x1 - 2 and (y2 == y1 + 1 or y2 == y1 - 1)) or \
   (x2 == x1 - 1 and (y2 == y1 + 2 or y2 == y1 - 2)) or \
   (x2 == x1 + 1 and (y2 == y1 + 2 or y2 == y1 - 2)) or \
   (x2 == x1 + 2 and (y2 == y1 + 1 or y2 == y1 - 1)):
    print("YES")
else:
    print("NO")
```

students/biegon_piotr/lesson_02_flow_control/lost_card.py
```diff
@@ -0,0 +1,17 @@
...
+N = int(input("Enter the number of cards: "))
+
+x = 0
+y = 0
+
+for i in range(1, N):
+    n = int(input("Enter the value of the card: "))
+    x += n
+
+for i in range(1, N + 1):
```
kuszcjan: Instead of iterating again over a range of values could combine those two `for` loops into single one 
or
use built-in sum function e.g.`y = sum(range(N+1))`

students/biegon_piotr/lesson_02_flow_control/the_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,16 @@
+print("The number of elements equal to the maximum\n")
+
+max1 = 0
```
kuszcjan: it's good that using `max` as name of variable was avoided (so no shadowing of built-ins).
However for sake of readability it would be good to use more elaborate names. In this scenario something like
`max_value` would fit better.

students/biegon_piotr/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,11 @@
+print("The number of zeros\n")
+
+N = int(input("Enter the number of numbers: "))
+result = 0
```
kuszcjan: should rename `result` to `counter` to reflect more it's purpose

kuszcjan:
Good!

PR #149, opened 2018-03-06
Finished lesson_01

wachumar:
@kuszcjan

### DIFF
--- a/students/wachulec_maria/lesson_01_basics/first_digit_after_decimal_point.py
+++ b/students/wachulec_maria/lesson_01_basics/first_digit_after_decimal_point.py
@@ -1,0 +1,6 @@
+from math import floor
+
+number = float(input("Real number: "))
+result = number - floor(number)
+print("First digit to the right of the decimal point is"
+      ": " + str(result)[2])
--- a/students/wachulec_maria/lesson_01_basics/hello_harry.py
+++ b/students/wachulec_maria/lesson_01_basics/hello_harry.py
@@ -1,0 +1,2 @@
+name = input("What is your name?: ")
+print("Hello, " + name + '.')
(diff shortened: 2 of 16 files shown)


students/wachulec_maria/lesson_01_basics/first_digit_after_decimal_point.py
```diff
@@ -0,0 +1,6 @@
+from math import floor
+
+number = float(input("Real number: "))
+result = number - floor(number)
+print("First digit to the right of the decimal point is"
+      ": " + str(result)[2])
```
kuszcjan: no newline at the end of file

students/wachulec_maria/lesson_01_basics/hello_harry.py
```diff
@@ -0,0 +1,2 @@
+name = input("What is your name?: ")
+print("Hello, " + name + '.')
```
kuszcjan: There should be an exclamation mark at the end.
This code does not pass any of tests on snakify website https://snakify.org/lessons/print_input_numbers/problems/hello_harry/
It's a good idea to run tests before submitting code.

students/wachulec_maria/lesson_01_basics/sum_of_digits.py
```diff
@@ -0,0 +1,5 @@
+number = input("Number: ")
+digits_sum = 0
+for i in number:
+    digits_sum = digits_sum + int(i)
```
kuszcjan: could use `+=` operator here

students/wachulec_maria/lesson_01_basics/tens_digit.py
```diff
@@ -0,0 +1,3 @@
+number = input("Give me number: ")
+number = '00' + number
+print(number[-2])
```
kuszcjan: Relying so much on slicing strings is not the best idea. It would be preferred to do some operations on integers.

PR #163, opened 2018-03-07
Lesson 3 solutions

synulewar:
@kuszcjan

### DIFF
--- a/students/synowiec_krzysztof/lesson_03_functions/calculator.py
+++ b/students/synowiec_krzysztof/lesson_03_functions/calculator.py
@@ -1,0 +1,94 @@
+def main():
+    welcome()
+    while True:
+        print("Provide option: ")
+        option = input()
+        if option == "a":
+            add()
+        elif option == "m":
+            multiply()
+        elif option == "s":
+            substract()
+        elif option == "d":
+            divide()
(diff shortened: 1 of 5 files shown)


students/synowiec_krzysztof/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,94 @@
+def main():
+    welcome()
```
kuszcjan: No docstring. example:
`"""Main entry point of the program"""`

students/synowiec_krzysztof/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,94 @@
...
+        elif option == "?" or option == "h":
+            help()
+        elif option == "q":
+            print("GOOD BYE")
+            break
+        else:
+            print("Option not recognized!")
+
+def getInput():
+    """Returns two integers based on user input
```
kuszcjan: If applicable, try to fit docstring in single line 
Check out https://www.python.org/dev/peps/pep-0257/#one-line-docstrings
e.g:
```
def getInput():
    """Returns two integers based on user input """
    print("Input 1st operand:")
```

PR #188, opened 2018-03-12
Snakify scripts for lesson_02_flow_control

glennbaker:
@tomekDiakowTieto


students/baker_glenn/snakify_lesson_3/bishop_moves.py
```diff
@@ -0,0 +1,15 @@
+# This program takes a bishops current position on the chessboard, and determines
+#if it is legal to move to another given position on the chessboard
+print("enter current position A")
+currentA = int(input())
```
tomekDiakowTieto: In python we should use underscore: current_a

students/baker_glenn/snakify_lesson_3/leap_year.py
```diff
@@ -0,0 +1,8 @@
+# script to determine if a year is a leap year or not
+print("enter the year")
+year = int(input())
+
+if (((year % 4) == 0) and ((year % 100) != 0) or ((year % 400) == 0)):
```
tomekDiakowTieto: Good

students/baker_glenn/snakify_lesson_4/factorial.py
```diff
@@ -0,0 +1,8 @@
+# script to calculate the factorial
+print("enter a number")
+number = int(input())
+
+result = 1
+for i in range(number):
```
tomekDiakowTieto: You can change range(number) to range(1, number + 1)  and result *= (i+1) on result * = i

PR #189, opened 2018-03-12
Gradopaw/lesson 01 basics

chylamal:


students/chylak_malgorzata/lesson_01_basics/hello_harry.py
```diff
@@ -0,0 +1,2 @@
+myName = input()
+print("Hello, " + myName)
```
gradopaw: Missing of the exclamation sign

students/chylak_malgorzata/lesson_01_basics/hello_harry.py
```diff
@@ -0,0 +1,2 @@
+myName = input()
```
gradopaw: Use underscores in the variables names: my_name

students/chylak_malgorzata/lesson_01_basics/fractional_part.py
```diff
@@ -0,0 +1,7 @@
+import math
+
+n = float(input())  # float, because there's a number with a comma
+print(math.modf(n))
+print('0.'.join(str(n).split('.')[1]))
+frac = n - int(n)
+print("%.2f" % frac)
```
gradopaw: Which of this three prints should I treat as a proper answer :)? The last line is printing almost good result. Change that print to print(frac) and comment or delete rest of the prints

PR #193, opened 2018-03-12
Lesson 2: flow control

hatan:
@kamwar please check it :)

### DIFF
--- a/students/kosarzewski_maciej/lesson_02_flow_control/chocolate_bar.py
+++ b/students/kosarzewski_maciej/lesson_02_flow_control/chocolate_bar.py
@@ -1,0 +1,9 @@
+x = int(input())
+y = int(input())
+part = int(input())
+surface = x*y
+if surface >= part and (surface-part == x or surface - part == y or x == part or y == part or
+                        (surface-part) % x == 0 or (surface-part) % y == 0):
+    print("YES")
+else:
+    print("NO")
--- a/students/kosarzewski_maciej/lesson_02_flow_control/chocolate_bar.py
+++ b/students/kosarzewski_maciej/lesson_02_flow_control/chocolate_bar.py
@@ -2,8 +2,7 @@ x = int(input())
 y = int(input())
 part = int(input())
 surface = x*y
-if surface >= part and (surface-part == x or surface - part == y or x == part or y == part or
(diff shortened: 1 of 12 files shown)


students/kosarzewski_maciej/lesson_02_flow_control/chocolate_bar.py
```diff
@@ -0,0 +1,8 @@
+x = int(input())
+y = int(input())
+part = int(input())
+surface = x*y
```
kamwar: It's not stated in PEP8 but generally it's recommended to use space between math operators
'surface = x * y'

students/kosarzewski_maciej/lesson_02_flow_control/the_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,10 @@
+number = int(input())
+biggest = 0
+numbers_list = []
+# Print a value:
+while number != 0:
+    if number >= biggest:
+        biggest = number
+        numbers_list.append(number)
+    number = int(input())
+print(numbers_list.count(biggest))
```
kamwar: Improvement proposal to the code above

```
biggest = 0
numbers_list = []
# Print a value:
while True:
    # It's better to write similary functionality like input statement in one place
    number = int(input())
    if not number:
        break
    elif number > biggest:
        biggest = number
        # number list should be consistent i.e only to include max values
        numbers_list = []
    if number >= biggest:
        numbers_list.append(number)
print(len(numbers_list))
```

students/kosarzewski_maciej/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,7 @@
+total = 0
+N = int(input())
```
kamwar: Please use lower case for variable names

students/kosarzewski_maciej/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,7 @@
...
+for i in range(0, N):
+    number = int(input())
+    if number == 0:
+        total += 1
+print (total)
```
kamwar: There should be no space after print statement

PR #195, opened 2018-03-12
Lesson 3 The Collatz sequence

synulewar:
@kuszcjan


students/synowiec_krzysztof/lesson_03_functions/the_collatz_sequence.py
```diff
@@ -0,0 +1,30 @@
...
+        except ValueError:
+            print("You must provide integer. Try again!")
+
+
+def collatz(number):
+    if number % 2 == 0:
+        result = (number // 2)
+    else:
+        result = 3 * number + 1
+    print (result)
```
kuszcjan: There should be no whitespace before `(`

PR #197, opened 2018-03-12
Solutions of the problems from lesson 2

BartoszPi:
@gradopaw please review


students/pietrewicz_bartosz/lesson_02_flow_control/Lost_card.py
```diff
@@ -0,0 +1,10 @@
+# Read the number of cards
+N = int(input())
+# Read the numbers on cards and calculate their sum
+sum = 0
```
gradopaw: Shadows built-in name 'sum'. Next time please avoid these names: https://docs.python.org/3/library/functions.html

PR #199, opened 2018-03-13
Kamwar/lesson 03 functions

swietmon:
@kamwar Please check

### DIFF
--- a/students/swietczak_monika/lesson_03_functions/calculator.py
+++ b/students/swietczak_monika/lesson_03_functions/calculator.py
@@ -1,0 +1,61 @@
+def calculate(option, a, b):
+    """
+    Performs calculations according to selected option
+    :param option: string, one of selected options
+    :param a: integer
+    :param b: integer
+    :return: number
+    """
+    if option == "a":
+        return a + b
+    if option == "s":
+        return a - b
+    if option == "m":
+        return a * b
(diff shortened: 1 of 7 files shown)


students/swietczak_monika/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Performs calculations according to selected option
+    :param option: string, one of selected options
+    :param a: integer
+    :param b: integer
+    :return: number
+    """
+    if option == "a":
+        return a + b
+    if option == "s":
```
kamwar: For code readability it's better to use elif here and handle default case with else when incorrect option was provided

students/swietczak_monika/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,61 @@
...
+def calculator(option):
+    """
+    Main function displaying info/result/help to user. It also runs calculate
+    function if user selects the right option.
+    :param option: String - selected option, one of values: a,s,m,d,p,h,?,q
+    """
+    try:
+        if option == "a":
+            print("ADDING")
+        if option == "s":
```
kamwar: For code readability it's better to use elif here and handle default case with else when incorrect option was provided

PR #201, opened 2018-03-13
Clarify first task for lesson 04

kuszcjan:
Some folks have problem with mocking input()

### DIFF
--- a/course/lesson_04_unit_testing/README.md
+++ b/course/lesson_04_unit_testing/README.md
@@ -5,10 +5,10 @@
 
 ### practice projects
 
-1. Modify (The Collatz Sequence) so that its main function takes an argument and return list with all numbers.
-Write `unittest.TestCase` tests (2-3) for it.
-(Optional) Replace `input()` function for tests basing on https://gist.github.com/kuszcjan/fcb89621191e28dad6f8263f1960e5c8 
-
+1. Using `unittest.TestCase` write tests for **The Collatz Sequence** exercise from lesson 3:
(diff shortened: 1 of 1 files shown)


jedzej:
To do what is requested they'd need to construct a list, add elements to it, iterate over it to print output in main and use list related assertions in test case. They don't know how to do it so far, since lists will be introduced in the next lesson. My proposal is to keep this this lesson out of it, but **move this exercise to next lesson** as very good example of tests improvements by involving richer python features.

In this lesson I'd focus on testing collatz function mainly. Let them pick representative test data covering typical cases, corner cases, invalid data etc. What do you think?

jedzej:
@kuszcjan I've forked and extended your snippet to additionally catch print: https://gist.github.com/jedzej/c9129558e87ec5d0b37405afd0ce6f46
With this it should be possible for them to test collatz without understanding lists:)

course/lesson_04_unit_testing/README.md
```diff
@@ -5,10 +5,11 @@
 
 ### practice projects
 
-1. Write unittest.TestCase tests (2-3) for (The Collatz Sequence) function. 
-Define test data for tests in setUp method. 
+1. Modify (The Collatz Sequence) so that its main function takes an argument and return list with all numbers.
```
adziu: add "from lesson 03"

course/lesson_04_unit_testing/README.md
```diff
@@ -5,10 +5,11 @@
...
+Write `unittest.TestCase` tests (2-3) for it.
```
adziu: This is supposed to mean 2 or 3? Give a specific test to implement, so that it's clear how it should be done. For example:
* test that it crashes on input 'not a number'
* test that for input 4 it retuns 2,

kuszcjan:
Removed monkey patching part. Need to merge this one or #225

PR #202, opened 2018-03-13
add lesson02 files

agatastachowska:
@kamwar lesson02 ready to check


students/stachowska_agata/lesson_02_flow_control/second_maximum.py
```diff
@@ -0,0 +1,13 @@
+second_max = 0
+first_max = 0
+while True:
+    number = int(input())
+    if number != 0:
```
kamwar: I prefer to reduce unnecessary else statement 
```
if not number:
    break
```

PR #203, opened 2018-03-13
Lesson 02 Flow control done

olszebar:
@kamwar  Please check lesson02

### DIFF
--- a/students/olszewski_bartosz/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
+++ b/students/olszewski_bartosz/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
@@ -1,0 +1,12 @@
+biggest = 0
+sum_number = 0
+while True:
+    n = int(input())
+    if n > biggest:
+        sum_number = 1
+        biggest = n
+    elif biggest == n:
+        sum_number += 1
+    if n == 0:
+        break
+print(sum_number)
--- a/students/olszewski_bartosz/lesson_02_flow_control/knight_move.py
+++ b/students/olszewski_bartosz/lesson_02_flow_control/knight_move.py
@@ -1,0 +1,10 @@
+base_x = int(input())
+base_y = int(input())
+direct_x = int(input())
+direct_y = int(input())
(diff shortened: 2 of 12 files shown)


students/olszewski_bartosz/lesson_02_flow_control/The_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,12 @@
+biggest = 0
+sum_number = 0
+while True:
+    n = int(input())
+    if n > biggest:
+        sum_number = 1
+        biggest = n
+    elif biggest == n:
+        sum_number += 1
+    if n == 0:
```
kamwar: There is no reason to analyze the section above 'if n == 0' for te last 0 element so for the code readability  'if n == 0' should be placed just below 'while'

```
while True:
    n = int(input())
    if n == 0:
        break
    elif n > biggest:
        sum_number = 1
        biggest = n
    elif n == biggest:
        sum_number += 1
```

students/olszewski_bartosz/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,10 @@
+base_x = int(input())
+base_y = int(input())
+direct_x = int(input())
+direct_y = int(input())
+if (abs(direct_x - base_x) == 2) and (abs(direct_y - base_y) == 1):
+    print('YES')
+elif (abs(direct_x - base_x) == 1) and (abs(direct_y - base_y) == 2):
```
kamwar: It might be compressed to one if  statement
if (abs(direct_x - base_x) == 2) and (abs(direct_y - base_y) == 1) or \
    (abs(direct_x - base_x) == 1) and (abs(direct_y - base_y) == 2):
    print('YES')

students/olszewski_bartosz/lesson_02_flow_control/ladder.py
```diff
@@ -0,0 +1,5 @@
+ladder = int(input())
+for j in range(1, ladder + 1):
+    for i in range(1, j + 1):
+        print(i, end='')
+    print(sep='\n')
```
kamwar: By default print uses sep='\n' so you might just print empty string: print('')

PR #226, opened 2018-03-14
Tomek diakow tieto/lesson 03 functions

AxelNowaczyk:
@tomekDiakowTieto

### DIFF
--- a/students/axel_nowaczyk/lesson_03_functions/calculator.py
+++ b/students/axel_nowaczyk/lesson_03_functions/calculator.py
@@ -1,0 +1,82 @@
+print("Welcome to badly organized calculator:")
+print("a - add")
+print("s - subtract")
+print("m - multiply")
+print("d - divide")
+print("p - power")
+print("h,? - help")
+print("q - QUIT")
+
+while True:
+    print("Enter option:")
+
+    option = input()
+
+    if option == "a":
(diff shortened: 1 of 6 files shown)


students/axel_nowaczyk/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,87 @@
...
+    print("POWER")
+    print("Input 1st operand:")
+    add_var_1 = int(input())
+    print("Input 2nd operand:")
+    add_var_2 = int(input())
+    print("Result:")
+    print(add_var_1 ** add_var_2)
+
+
+def start_calculator():
```
tomekDiakowTieto: You can use sth like:
options = {
   'a': parform_addition
   ....
}
 and later 
options[option]()

but your solution is good too

students/axel_nowaczyk/lesson_03_functions/fibonacci_numbers.py
```diff
@@ -0,0 +1,16 @@
+def fib(number):
+    if number <= 0:
+        return 0
+    elif number == 1:
+        return 1
+    else:
+        return fib(number - 1) + fib(number - 2)
```
tomekDiakowTieto: Good

PR #228, opened 2018-03-14
Lesson_03

piotrbiegon:
@kuszcjan

### DIFF
--- a/students/biegon_piotr/lesson_03_functions/calculator.py
+++ b/students/biegon_piotr/lesson_03_functions/calculator.py
@@ -1,0 +1,56 @@
+print("Welcome to the awesome calculator :)")
+
+
+def options():
+    print("a - add")
+    print("s - subtract")
+    print("m - multiply")
+    print("d - divide")
+    print("p - power")
+    print("h, ? - help")
+    print("q - QUIT")
+
+
+def data():
+    var_1 = float(input("Enter the first number: "))
+    var_2 = float(input("Enter the second number: "))
+    return var_1, var_2
+
+
+while True:
+    print("\nAvailable options:")
+    options()
+    option = input("\nChoose one option: ")
+
+    if option == "a":
+        print("\nADDING")
+        var_1, var_2 = data()
+        print("Addition result: {:f}".format(var_1 + var_2))
+
+    if option == "s":
+        print("\nSUBTRACT")
+        var_1, var_2 = data()
+        print("Subtraction result: {:f}".format(var_1 - var_2))
+
+    if option == "m":
+        print("\nMULTIPLY")
+        var_1, var_2 = data()
+        print("Multiplication result: {:f}".format(var_1 * var_2))
+
+    if option == "d":
+        print("\nDIVIDE")
+        var_1, var_2 = data()
+        print("Division result: {:f}".format(var_1 / var_2))
+
+    if option == "p":
+        print("\nPOWER")
+        var_1, var_2 = data()
+        print("The result of exponentiation: {:f}".format(var_1 ** var_2))
+
+    if option == "h" or option == "?":
+        print("\nHELP")
+
+    if option == "q":
+        break
+
+print("GOOD BYE")
--- a/students/biegon_piotr/lesson_03_functions/exponentiation.py
+++ b/students/biegon_piotr/lesson_03_functions/exponentiation.py
@@ -1,0 +1,14 @@
+print("Exponentiation\n")
+
+
+def power(a, n):
+    if n == 0:
+        return 1
+    if n >= 1:
+        return a * power(a, n-1)
+
+
+a = float(input("Enter a power base: "))
+n = int(input("Enter an exponent exponent: "))
+
+print("\nThe result is: {:f}".format(power(a, n)))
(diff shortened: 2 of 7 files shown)


students/biegon_piotr/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def data():
+    var_1 = float(input("Enter the first number: "))
+    var_2 = float(input("Enter the second number: "))
+    return var_1, var_2
+
+
+while True:
+    print("\nAvailable options:")
```
kuszcjan: I would move this line into `options()` method

students/biegon_piotr/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,56 @@
...
+    options()
+    option = input("\nChoose one option: ")
+
+    if option == "a":
+        print("\nADDING")
+        var_1, var_2 = data()
+        print("Addition result: {:f}".format(var_1 + var_2))
```
kuszcjan: The idea was to put every action into separate method.
So in the main loop would like more like this:
```
       if option == "a":
            add()
```

students/biegon_piotr/lesson_03_functions/exponentiation.py
```diff
@@ -0,0 +1,14 @@
+print("Exponentiation\n")
+
+
+def power(a, n):
+    if n == 0:
+        return 1
+    if n >= 1:
```
kuszcjan: I think here `elif` would suit better here.
There so no way that both conditions gonna ever apply, so to empathize that `elif` is more preferred.

students/biegon_piotr/lesson_03_functions/input_validation.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+while 1:
+    try:
+        number = int(input("Enter a number: "))
+        if collatz(number) != 1:
+            print(collatz(number))
+        else:
+            break
+    except:
```
kuszcjan: it's too broad exception clause. We should not catch all exceptions, only those we can handle. In this scenario, we are looking for `ValueError` exception

students/biegon_piotr/lesson_03_functions/input_validation.py
```diff
@@ -0,0 +1,21 @@
...
+    if number % 2 == 0:
+        return number // 2
+    else:
+        return 3 * number + 1
+
+
+while 1:
+    try:
+        number = int(input("Enter a number: "))
+        if collatz(number) != 1:
```
kuszcjan: in `try...except` block we should not put much code. the less the better. Would `print(collatz(number))` raise an ValueError exception ?

students/biegon_piotr/lesson_03_functions/input_validation.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def collatz(number):
+    if number % 2 == 0:
+        return number // 2
+    else:
+        return 3 * number + 1
+
+
+while 1:
```
kuszcjan: Logic is wrong. This code asks user every time to get input, where the task is different:
> Then write a program that lets the user type in an integer and **that keeps calling collatz() on that number until the function returns the value 1.**

So for single input, a bunch of numbers should be printed, until `1` is reached.

students/biegon_piotr/lesson_03_functions/input_validation.py
kuszcjan: Take a look at  `try...except...elese` clause.
From https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python:

> Without the else-clause, the only option to run additional code before finalization would be the **clumsy practice of adding the code to the try-clause**. That is clumsy because it risks raising exceptions in code that wasn't intended to be protected by the try-block.

students/biegon_piotr/lesson_03_functions/the_collatz_sequence.py
```diff
@@ -0,0 +1,18 @@
+print("The Collatz Sequence\n")
```
kuszcjan: don't need to commit collatz_sequence, when there is better version with input validation :)

students/biegon_piotr/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,82 @@
+print("Welcome to the awesome calculator :)")
```
kuszcjan: In previous review forgot to mention that there is no docstrings in whole file.

Every method should have docstring.
https://www.python.org/dev/peps/pep-0257/

students/biegon_piotr/lesson_03_functions/input_validation.py
```diff
@@ -0,0 +1,21 @@
...
+
+def collatz(number):
+    if number % 2 == 0:
+        return number // 2
+    else:
+        return 3 * number + 1
+
+
+def loop():
+    x = int(input("Enter a number: "))
```
kuszcjan: this will work, but ideally `try...except` should be just around line 12

PR #229, opened 2018-03-14
Lesson 2 completed

wachumar:
@kuszcjan

### DIFF
--- a/students/wachulec_maria/lesson_02_flow_control/knight_move.py
+++ b/students/wachulec_maria/lesson_02_flow_control/knight_move.py
@@ -1,0 +1,12 @@
+start_column = int(input("Start column: "))
+start_row = int(input("Start row: "))
+end_column = int(input("End column: "))
+end_row = int(input("End row: "))
+
+delta_columns = abs(start_column - end_column)
+delta_rows = abs(start_row - end_row)
+
+if (delta_columns == 1 and delta_rows == 2) or (delta_columns == 2 and delta_rows == 1):
+    print("YES")
+else:
+    print("NO")
--- a/students/wachulec_maria/lesson_02_flow_control/the_number_of_elements_equal_to_the_maximum.py
+++ b/students/wachulec_maria/lesson_02_flow_control/the_number_of_elements_equal_to_the_maximum.py
@@ -1,0 +1,13 @@
+number = int(input("Number: "))
+large_number = number
+sum_of_large_number = 1
+while True:
+    number = int(input("Number: "))
+    if number > large_number:
+        large_number = number
+        sum_of_large_number = 1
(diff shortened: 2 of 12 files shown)


students/wachulec_maria/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,12 @@
+start_column = int(input("Start column: "))
+start_row = int(input("Start row: "))
+end_column = int(input("End column: "))
+end_row = int(input("End row: "))
+
+delta_columns = abs(start_column - end_column)
+delta_rows = abs(start_row - end_row)
+
+if (delta_columns == 1 and delta_rows == 2) or (delta_columns == 2 and delta_rows == 1):
```
kuszcjan: although not required, that parenthesis improve readability. nice

students/wachulec_maria/lesson_02_flow_control/the_number_of_elements_equal_to_the_maximum.py
```diff
@@ -0,0 +1,13 @@
...
+while True:
+    number = int(input("Number: "))
+    if number > large_number:
+        large_number = number
+        sum_of_large_number = 1
+    elif number == large_number:
+        sum_of_large_number += 1
+    if number == 0:
+        print(sum_of_large_number)
+        break
```
kuszcjan: it would be better not to use `break` and put a condition to while loop .e.g: `while n != 0:`

students/wachulec_maria/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,15 @@
+base_1 = int(input("Number: "))
+base_2 = int(input("Number: "))
```
kuszcjan: Little bit too foggy var names. it's better to aim in some clear var names like `second_max, first_max`

students/wachulec_maria/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,15 @@
...
+while True:
+    number = int(input("Number: "))
+    if base_2 > base_1 and number > base_1:
```
kuszcjan: logic could be simplified, but its okay

students/wachulec_maria/lesson_02_flow_control/knight_move.py
kuszcjan: It's also worth noting that complex `if` statement can be transformed to multiline form.
https://www.python.org/dev/peps/pep-0008/#indentation
https://stackoverflow.com/questions/181530/styling-multi-line-conditions-in-if-statements

PR #233, opened 2018-03-14
tasks for lesson_03_functions

miroslawj:

### DIFF
--- a/students/jakielczyk_miroslaw/lesson_03_functions/Collatz_Sequence.py
+++ b/students/jakielczyk_miroslaw/lesson_03_functions/Collatz_Sequence.py
@@ -1,0 +1,16 @@
+# Collatz Sequence
+print ("Enter the number ")
+number = int(input())
+def collatz(number):
+    if number % 2 == 0:
+        print (number // 2)
+        return (number // 2)
+    elif number % 2 == 1:
+        print (3 * number + 1)
+        return (3 * number + 1)
+result = number
+while result != 1:
+    result = collatz(result)
+
+        
+        
--- a/students/jakielczyk_miroslaw/lesson_03_functions/Collatz_Sequence.py
+++ b/students/jakielczyk_miroslaw/lesson_03_functions/Collatz_Sequence.py
@@ -3,14 +3,13 @@ print ("Enter the number ")
 number = int(input())
 def collatz(number):
     if number % 2 == 0:
-        print (number // 2)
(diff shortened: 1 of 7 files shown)


students/jakielczyk_miroslaw/lesson_03_functions/Collatz_Sequence.py
```diff
@@ -0,0 +1,16 @@
+# Collatz Sequence
+print ("Enter the number ")
+number = int(input())
+def collatz(number):
+    if number % 2 == 0:
+        print (number // 2)
+        return (number // 2)
+    elif number % 2 == 1:
+        print (3 * number + 1)
```
kuszcjan: it would be better to have single `print` outside of if construct.
Or even better to move it to while loop and use `print(result)`

students/jakielczyk_miroslaw/lesson_03_functions/Exponentiation.py
```diff
@@ -0,0 +1,9 @@
+# Exponentiation
+a = float(input())
+n = int(input())
+def power (a, n):
```
kuszcjan: there should be blank line before function definition
> Surround top-level function and class definitions with two blank lines.
Method definitions inside a class are surrounded by a single blank line.

https://www.python.org/dev/peps/pep-0008/#blank-lines

students/jakielczyk_miroslaw/lesson_03_functions/kalkulator.py
```diff
@@ -0,0 +1,57 @@
...
+    print("q - QUIT")
+
+def provide_inputs():
+    print("Input 1st operand:")
+    add_var_1 = int(input())
+    print("Input 2nd operand:")
+    add_var_2 = int(input())
+    return add_var_1, add_var_2
+    
+def add (inputs):
```
kuszcjan: do not add space before '('

students/jakielczyk_miroslaw/lesson_03_functions/kalkulator.py
```diff
@@ -0,0 +1,57 @@
+import math
```
kuszcjan: There is no docstrings in the whole file. Please add them.

kuszcjan: https://www.python.org/dev/peps/pep-0257/

PR #237, opened 2018-03-15
Adziu/lesson_02_flow_control

marioosh11:
@adziu please review


students/mariusz_michalczyk/lesson_02_flow_control/The_number_of_elements_equa_ to_the_maximum.py
```diff
@@ -0,0 +1,13 @@
+# NOT FINISHED
+
+equal_nrs, largest, i = 0, 0, 0
```
adziu: equal_nrs starts as `1`

students/mariusz_michalczyk/lesson_02_flow_control/The_number_of_zeros.py
```diff
@@ -0,0 +1,10 @@
+size = int(input("How many numbers you want to enter: "))
+zeros = 0
+id = (size+1) - size
+while size:
```
adziu: Typically, we write `for index in range(1, size + 1)`, also `id` is a built-in function that gives you the id of an object (typically memory pointer).

PR #239, opened 2018-03-15
increase-max-line-length

kuszcjan:
Proposal of increasing max line length to some more sane value.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,5 +1,5 @@
 linters:
   flake8:
     ignore: 'E999'
-    max-line-length: 79
+    max-line-length: 120
     fixer: false


adziu:
I'm against.

jedzej:
I'm against too. These programs are simple enough to follow this restriction, so it shouldn't be a problem for participants to fit in. The benefits are:
 - keeping 79 sings per line learns them when and how they should break the lines
 - it's easy for us to use side-by-side compares when reviewing code on almost every screen

zamber:
I'm for.
Side by side comparison is doable as well with 120 chars.
Breaking lines is more of a hassle than profit. PEP8 is pretty nitpicky about that.
It's 2018, not 1988. We have screens that can accommodate 120 chars.

Problems arise for people when they pass longish strings to `input()`. They usually go over 79 by 2-6 chars.

We will get more complaints about that than valuable feedback. If they go over 120 then there's an actual issue. 79 is just a PITA that some over-opinionated `stickler` bot blocks them on without any sensible explanation.

adziu:
If you think it's about 2-6 chars, then go for 100 instead of 120. 

[trigger warning: I call lazy people names :grin: ]
It's not about screens, it's about human cognition. Find me a good book, that was not done in Microsoft Word, with 120 characters on a line. I don't buy this argument, it's just being lazy about readability.

kuszcjan:
What is readable and what is not differ very much from person to person. 
For this course, I think a good compromise would be around mentioned 100.

jedzej:
100 is fine for me

PR #242, opened 2018-03-15
Gradopaw/lesson 02

akpoik:
@gradopaw


akpoik:
@gradopaw

students/kulon_aleksander/lesson_02/knight_move.py
```diff
@@ -0,0 +1,12 @@
+cell1_x = int(input())
+cell1_y = int(input())
+
+cell2_x = int(input())
+cell2_y = int(input())
+
+if ((cell1_y + 2 == cell2_y) or (cell1_y - 2 == cell2_y)) and ((cell1_x + 1 == cell2_x) or (cell1_x - 1 == cell2_x)):
```
gradopaw: A bit complicated. This and next "if" could be simplified, so instead two "if" statements you can have something like this:
if abs(cell1_x - cell2_x) == 2 and abs(cell1_y - cell2_y) == 1 or \
       abs(cell1_y - cell2_y) == 2 and abs(cell1_x - cell2_x) == 1:

In general results you got are ok so no need to fix it.

students/kulon_aleksander/lesson_02/lost_card.py
```diff
@@ -0,0 +1,9 @@
+n = int(input())
+result = 0
+
+for i in range(1, n + 1):
+    result += i
+
+for i in range(1, n):
+    result -= int(input())
+print(result)
```
gradopaw: To have code more readable, before print it is good for have one blank line.

students/kulon_aleksander/lesson_02/queen_move.py
```diff
@@ -0,0 +1,18 @@
+cell1_x = int(input())
```
gradopaw: Have a look at:

cell1_x = int(input())
cell1_y = int(input())

cell2_x = int(input())
cell2_y = int(input())

if abs(cell1_x - cell2_x) == abs(cell1_y - cell2_y) or \
           cell1_x == cell2_x or cell1_y == cell2_y:
         print('YES')
else:
        print('NO')

akpoik:
@gradopaw

students/kulon_aleksander/lesson_02/lost_card.py
```diff
@@ -1,9 +1,10 @@
 n = int(input())
-result = 0
 
+sum = 0
```
gradopaw: Why did you change this name :) sum is built-in function - try to avoid built-in functions names:
https://docs.python.org/3/library/functions.html

PR #244, opened 2018-03-15
Lesson 02 - 12 tasks

karinanartowska:
@gradopaw Please review :)

### DIFF
--- a/students/nartowska_karina/lesson_02_flow_control/factorial.py
+++ b/students/nartowska_karina/lesson_02_flow_control/factorial.py
@@ -1,0 +1,6 @@
+value = int(input())
+i = 1
+tmp = 1
+for i in range(2, value + 1):
+    tmp = tmp * i
+print(tmp)
--- a/students/nartowska_karina/lesson_02_flow_control/knight_move.py
+++ b/students/nartowska_karina/lesson_02_flow_control/knight_move.py
@@ -2,7 +2,7 @@ c1 = int(input())
 r1 = int(input())
 c2 = int(input())
 r2 = int(input())
-if (abs(r1 - r2) == 1 and abs(c1 - c2) == 2): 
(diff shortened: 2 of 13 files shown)


students/nartowska_karina/lesson_02_flow_control/factorial.py
```diff
@@ -0,0 +1,6 @@
+value = int(input())
+i = 1
```
gradopaw: There is no need to set an initial value for "i" variable in a case when it is used as an iterator in the "for" loop.

students/nartowska_karina/lesson_02_flow_control/lost_card.py
```diff
@@ -0,0 +1,9 @@
+value = int(input())
+tmp = 0
+i = 0
```
gradopaw: The same as previous - not needed

students/nartowska_karina/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,10 @@
+c1 = int(input())
+r1 = int(input())
+c2 = int(input())
+r2 = int(input())
+if (abs(r1 - r2) == 1 and abs(c1 - c2) == 2):
```
gradopaw: Redundant parentheses:
if abs(r1 - r2) == 1 and abs(c1 - c2) == 2:

students/nartowska_karina/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,10 @@
+c1 = int(input())
+r1 = int(input())
+c2 = int(input())
+r2 = int(input())
+if (abs(r1 - r2) == 1 and abs(c1 - c2) == 2): 
+    print("YES")
+elif (abs(r1 - r2) == 2 and abs(c1 - c2) == 1):
```
gradopaw: As above

PR #263, opened 2018-03-19
add lesson 03

agatastachowska:
ready to check @kamwar

### DIFF
--- a/students/stachowska_agata/lesson_03_functions/calculator.py
+++ b/students/stachowska_agata/lesson_03_functions/calculator.py
@@ -49,34 +49,39 @@ while True:
         print("Result:")
         print(add(var_1, var_2))
 
-    if option == "s":
+    elif option == "s":
         print("SUBTRACT")
         var_1, var_2 = input_operand()
         print("Result:")
         print(subtrack(var_1, var_2))
 
-    if option == "m":
+    elif option == "m":
         print("MULTIPLY")
         var_1, var_2 = input_operand()
         print("Result:")
         print(multiply(var_1, var_2))
 
-    if option == "d":
+    elif option == "d":
         print("DIVIDE")
         var_1, var_2 = input_operand()
         print("Result:")
         print(divide(var_1, var_2))
 
-    if option == "p":
+    elif option == "p":
         print("POWER")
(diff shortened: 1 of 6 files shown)


students/stachowska_agata/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,82 @@
...
+    print("Enter option:")
+    option = input()
+
+    if option == "a":
+        print("ADDING")
+        var_1, var_2 = input_operand()
+        print("Result:")
+        print(add(var_1, var_2))
+
+    if option == "s":
```
kamwar: For code readability please use 'elif' instead of 'if' when possible

students/stachowska_agata/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,82 @@
...
+        print("Result:")
+        print(power(var_1, var_2))
+
+    if option == "h" or option == "?":
+        print("HELP")
+        help()
+
+    if option == "q":
+        print("GOOD BYE")
+        break
```
kamwar: Missing default else for unrecognized argument

students/stachowska_agata/lesson_03_functions/collatz_sequence.py
```diff
@@ -0,0 +1,18 @@
...
+    else:
+        res = 3 * x + 1
+    print(res)
+    return res
+
+
+while True:
+    try:
+        number = int(input("Enter a number: "))
+
```
kamwar: The exception handling should include only the part of code which might raise it

```
try:
    number = int(input("Enter a number: "))
except ValueError:
    print('You must enter an integer!')
    continue
```

students/stachowska_agata/lesson_03_functions/power_rec.py
```diff
@@ -0,0 +1,14 @@
+def power(a, n):
+    result = 1
+    while True:
+        if n == 0:
+            break
+        else:
```
kamwar: else is not necessary here
```
def power(a, n):
    result = 1
    while True:
        if n == 0:
            break
        result *= a
        n -= 1
    return result
```

PR #264, opened 2018-03-19
lesson02 tasks

jemiekam:


jemiekam:
@kamwar

students/jemielity_kamil/lesson_02_flow_control/adding_factorials.py
```diff
@@ -0,0 +1,10 @@
+number = int(input('Number: '))
+
+total = 1
+factorial_sum = 0
+for x in range(1, number + 1):
+    total = total * x
+    factorial_sum += total
+
+
```
kamwar: unnecessary new line

students/jemielity_kamil/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,16 @@
+
+maximum = 0
+second_maximum = 0
+
+while True:
+    n = int(input("Integer: "))
+    if n == 0:
```
kamwar: Usually in python we use convention:
'if not n' instead of 'n == 0'. It's more pythonic but on the other side it's also more generic, n might be in [0, None, '', ""]

PR #266, opened 2018-03-19
Lesson 3

wachumar:
@kuszcjan

### DIFF
--- a/students/wachulec_maria/lesson_03_functions/calculator.py
+++ b/students/wachulec_maria/lesson_03_functions/calculator.py
@@ -1,0 +1,106 @@
+def show_available_options():
+    """
+    Print available options
+    """
+    print("Welcome to not-badly organized calculator:")
+    print("a - add")
+    print("s - subtract")
+    print("m - multiply")
+    print("d - divide")
+    print("p - power")
+    print("h,? - help")
+    print("q - QUIT")
+
+
+def take_two_number():
+    """
+    Input two number and return them
+    """
+    first_number = int(input("Input 1st operand:\n"))
+    second_number = int(input("Input 2st operand:\n"))
+    return first_number, second_number
+
+
+def addition():
+    """
+    Do add operation between output of
+    take_two_number() function
+    """
+    print("ADDING")
+    x1, x2 = take_two_number()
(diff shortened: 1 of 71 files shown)


students/wachulec_maria/lesson_03_functions/the_length_of_the_segment.py
```diff
@@ -0,0 +1,10 @@
+def distance(x1, y1, x2, y2):
+    d = ((x2 - x1)**2 + (y2 - y1)**2)**(1/2)
```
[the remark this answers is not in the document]
kuszcjan: please fix that

students/wachulec_maria/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,74 @@
+def show_available_options():
```
kuszcjan: There is no docstrings in the whole file. Please add them.
https://www.python.org/dev/peps/pep-0257/

students/wachulec_maria/lesson_03_functions/the_collatz_sequence_and_input_validation.py
```diff
@@ -0,0 +1,15 @@
+def collatz(number):
+    if number % 2 == 0:
+        print(number // 2)
+        return number // 2
+    elif number % 2 == 1:
+        print(3 * number + 1)
```
kuszcjan: it would be better to move `print` after line 13
So there will single print after `collatz` function

students/wachulec_maria/lesson_03_functions/the_collatz_sequence_and_input_validation.py
```diff
@@ -0,0 +1,15 @@
...
+        return 3 * number + 1
+
+
+try:
+    n = int(input("Take me number: "))
```
kuszcjan: Don't mix `"` with `'` in the single file(project) for text. Be consistent. Use one convention across files.

In line 15 there is the single quote, where on line 11 there is a double quote.

kuszcjan:
There is something terribly wrong with this PR, need to check later.

kuszcjan:
BAD PR, closed

PR #270, opened 2018-03-20
Solutions for problems from lesson 03

BartoszPi:
@gradopaw please review

### DIFF
--- a/students/pietrewicz_bartosz/lesson_03_functions/the_collatz_sequence.py
+++ b/students/pietrewicz_bartosz/lesson_03_functions/the_collatz_sequence.py
@@ -1,0 +1,32 @@
+def collatz(number):
+    """Calculates and prints the next element of Collatz sequence"""
+    if number % 2 == 0:
+        number //= 2
+    else:
+        number = number * 3 + 1
+    print(number)
+    return number
+
+
+def read_number():
+    number = 0
+    # read integer from user until it is positive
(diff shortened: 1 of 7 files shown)


students/pietrewicz_bartosz/lesson_03_functions/the_collatz_sequence.py
```diff
@@ -0,0 +1,22 @@
...
+    if number % 2 == 0:
+        number //= 2
+    else:
+        number = number * 3 + 1
+    print(number)
+    return number
+
+
+number = 0
+# read integer from user until it is positive
```
gradopaw: Read integer from user could be in a separate function

students/pietrewicz_bartosz/lesson_03_functions/the_collatz_sequence.py
```diff
@@ -0,0 +1,22 @@
...
+while number < 1:
+    print("Enter positive integer: ", sep="", end="")
+    number = int(input())
+    if number < 1:
+        print("The number must be positive!")
+# calculate Collatz sequence until function calculating next element
+# returns 1
+number = collatz(number)
```
gradopaw: Is this is really needed? I think it's enough to have collatz() function call in the while loop

BartoszPi:
@gradopaw please review

PR #272, opened 2018-03-20
Adziu/lesson 03 functions

RCanDo:
@adziu


students/arkadiusz_kasprzyk/lesson_03_functions/badly_organized_calculator.py
```diff
@@ -0,0 +1,79 @@
...
+    print("a - add")
+    print("s - subtract")
+    print("m - multiply")
+    print("d - divide")
+    print("p - power")
+    print("h,? - help")
+    print("q - QUIT")
+
+
+def inputs(message):
```
adziu: `prompt_for_input` would be a better name and `message` is actually an `operand`, not a whole message.

students/arkadiusz_kasprzyk/lesson_03_functions/exponentiation.py
```diff
@@ -0,0 +1,18 @@
+def power(a, n):
+
+    if n < 0:
+        a = 1 / a
+        n = -n
+        return power(a, n)
```
adziu: Great idea.

PR #273, opened 2018-03-20
lesson_03_functions

olszebar:
@kamwar please check lesson_03_functions

### DIFF
--- a/students/olszewski_bartosz/lesson_03_functions/collatz.py
+++ b/students/olszewski_bartosz/lesson_03_functions/collatz.py
@@ -6,7 +6,7 @@ def collatz(number):
         print(3 * number + 1)
         return 3 * number + 1
 
-    
+
 try:
     a = int(input('Enter number: '))
     while a != 1:
--- a/students/olszewski_bartosz/lesson_03_functions/collatz.py
+++ b/students/olszewski_bartosz/lesson_03_functions/collatz.py
@@ -1,10 +1,10 @@
 def collatz(number):
     if number % 2 == 0:
-        print(number // 2)
-        return number // 2
-    elif number % 2 == 1:
-        print(3 * number + 1)
-        return 3 * number + 1
+        value = number // 2
+    else:
+        value = 3 * number + 1
+    print(value)
+    return value
 
 
 try:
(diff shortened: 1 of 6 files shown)


students/olszewski_bartosz/lesson_03_functions/exponentiation.py
```diff
@@ -0,0 +1,11 @@
+def exponentiation(a, n):
+    result = 1
+    if n == 0:
+        return 1
+    else:
```
kamwar: else is not necessary here

students/olszewski_bartosz/lesson_03_functions/collatz.py
```diff
@@ -0,0 +1,15 @@
+def collatz(number):
+    if number % 2 == 0:
+        print(number // 2)
+        return number // 2
+    elif number % 2 == 1:
```
kamwar: not necessary elif statement. When using if <-> elif always ends conditions with else (if <-> elif  <-> else) even if you think that it will never reach else.

students/olszewski_bartosz/lesson_03_functions/exponentiation.py
```diff
@@ -0,0 +1,8 @@
+def exponentiation(a, n):
+    if n == 0:
+        return 1
+    else:
+        return a * exponentiation(a, n - 1)
+        result *= a
```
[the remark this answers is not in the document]
kamwar: This statement is unreachable as it's after return

students/olszewski_bartosz/lesson_03_functions/negative_exponent.py
```diff
@@ -0,0 +1,13 @@
+def power(a, n):
+    result = 1
+    if n > 0:
+        for i in range(n):
+            result *= a
+    elif n < 0:
+        a = 1 / a
+        for i in range(abs(n)):
+            result *= a
```
kamwar: Try to avoid duplication of code

```
def power(a, n):
    result = 1
    for i in range(abs(n)):
        result *= a
    if n >= 0:
        return result
    else:
        return 1 / result
```

olszebar:
@kamwar  Please check it again

PR #274, opened 2018-03-20
Scripts for lesson 3

glennbaker:
@tomekDiakowTieto

### DIFF
--- a/students/baker_glenn/lesson_3_scripts/calculator.py
+++ b/students/baker_glenn/lesson_3_scripts/calculator.py
@@ -12,11 +12,13 @@ def get_input():
         except:
             print("please enter 2 integers")
 
+
 def addition():
 
     add_var_1, add_var_2 = get_input()
     print("Result:")
     print(add_var_1 + add_var_2)
+
 
 def subtract():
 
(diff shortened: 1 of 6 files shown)


students/baker_glenn/lesson_3_scripts/calculator.py
```diff
@@ -0,0 +1,105 @@
...
+    print("d - divide")
+    print("p - power")
+    print("h,? - help")
+    print("q - QUIT")
+
+
+print("Welcome to badly organized calculator:")
+print_options()
+
+while True:
```
tomekDiakowTieto: Tou can create sth like:

options = {
    'a': addition,
    ....
}

 and later options[option]()

but if-else are good soultion too

students/baker_glenn/lesson_3_scripts/collatz_sequence.py
```diff
@@ -0,0 +1,21 @@
...
+    else:
+        new_number = 3 * number + 1
+    if number == 1:
+        raise SystemExit
+    print(new_number)
+    collatz(new_number)
+
+
+print("enter an integer")
+while True:
```
tomekDiakowTieto: input_number = None
while input_number is None

PR #282, opened 2018-03-20
Lesson 4 solutions

synulewar:
@kuszcjan

### DIFF
--- a/students/synowiec_krzysztof/lesson_04_unit_testing/test_the_length_of_the_segment.py
+++ b/students/synowiec_krzysztof/lesson_04_unit_testing/test_the_length_of_the_segment.py
@@ -1,0 +1,33 @@
+import pytest
+from math import sqrt
+
+
+# Function under test
+def distance(x1, y1, x2, y2):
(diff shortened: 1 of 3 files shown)


students/synowiec_krzysztof/lesson_04_unit_testing/test_the_length_of_the_segment.py
```diff
@@ -0,0 +1,33 @@
...
+    return sqrt((x1 - x2) ** 2 + (y1 - y2) ** 2)
+
+
+def test_distance_none_parameters():
+    with pytest.raises(TypeError):
+        distance(None, None, None, None)
+
+
+def test_distance_string_parameters():
+    # Test that it crashes with ValueError when called on 'aoeu' - why?
```
kuszcjan: There will be ValueError when casting to float/ing would be inside function and it would raise an exception e.g.
`float('aoeu')`

kuszcjan:
Many small PEP8 violations.

PR #284, opened 2018-03-20
added lesson 3: functions

3tom:
@tomekDiakowTieto , please take a look.

Thanks!
Tomasz

### DIFF
--- a/students/medrek_tomasz/lesson_03_functions/calculator.py
+++ b/students/medrek_tomasz/lesson_03_functions/calculator.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+
+def calculate(a, b, option):
+    if option == 'a':
+        return a + b
+    elif option == 's':
+        return a - b
+    elif option == 'm':
+        return a * b
+    elif option == 'd':
+        return a / b
+    elif option == 'e':
+        return a ** b
+
+
+def help():
+    print('a - add\n'
+          's - subtract\n'
+          'm - multiply\n'
+          'd - divide\n'
+          'e - exponentiate\n'
+          'q - QUIT')
+
+
+options = ['a', 's', 'm', 'd', 'e', 'q']
+
+print('Welcome to calculator:')
+help()
+
+while True:
+    option = input('Please enter an option\n')
+    if option not in options:
+        print('Try again with possible option!\n')
+        help()
+        continue
+
+    if option == 'q':
+        print('GOOD BYE')
+        break
+
+    try:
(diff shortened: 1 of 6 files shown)


students/medrek_tomasz/lesson_03_functions/exponentation.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+
+
+def exponentiation(a, n):
```
tomekDiakowTieto: FYI:
return 1 if n == 0 else a * exponentiation(a, n - 1)

But propably my solution will be unreadable, so this is only TIP for You

students/medrek_tomasz/lesson_03_functions/fibonacci_numbers.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def fibonacci(n):
+    if (n == 0) or (n == 1):
+        return n
+    else:
+        return fibonacci(n - 1) + fibonacci(n - 2)
+
+
+try:
```
tomekDiakowTieto: You can put it inside while loop

students/medrek_tomasz/lesson_03_functions/exponentation.py
3tom: Nice tip, didn't know such thing is allowed, thanks.

students/medrek_tomasz/lesson_03_functions/fibonacci_numbers.py
3tom: https://snakify.org/lessons/functions/problems/fibonacci_rec/
"Don't use loops, use the flair of recursion instead. However, you should think about why the recursive method is much slower than using loops. "

3tom:
@tomekDiakowTieto , sorry I missed it somehow. Already uploaded this calculator, please check.
Thanks!

students/medrek_tomasz/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,49 @@
...
+        break
+
+    try:
+        x = float(input('Please enter first operand\n'))
+        y = float(input('Please enter second operand\n'))
+    except ValueError:
+        print('Not a valid, number, please try again\n')
+        continue
+
+    print('Result is:\n{}\n'.format(calculate(x, y, option)))
```
TomaszDiakowCoderslab: Looks great

3tom: It's old version o0
let me reupload

PR #289, opened 2018-03-21
flow_control_solutions

chylamal:


students/chylak_malgorzata/lesson_02_flow_control/adding_factorials.py
```diff
@@ -0,0 +1,7 @@
+n = int(input())
+factorial = 1
+sum = 0
```
gradopaw: Try to avoid using built-in functions. Please have a look:
https://docs.python.org/3.6/library/functions.html

students/chylak_malgorzata/lesson_02_flow_control/bishop_moves.py
```diff
@@ -0,0 +1,9 @@
+x1 = int(input())
+y1 = int(input())
+x2 = int(input())
+y2 = int(input())
+
+if(x1 - x2) == (y1 - y2):
```
gradopaw: You should compare absolute values:
if (abs(x1 - x2) == (y1 - y2):

students/chylak_malgorzata/lesson_02_flow_control/queen_move.py
```diff
@@ -0,0 +1,9 @@
+x1 = int(input())
+y1 = int(input())
+x2 = int(input())
+y2 = int(input())
+
+if x1 == x2 or y1 == y2:
```
gradopaw: Not all of the tests passed on Snakify.You should check as well following condition:
(abs(x1 - x2) == abs(y1 - y2)

so if statement should looks:
if x1 == x2 or y1 == y2 or abs(x1 - x2) == abs(y1 - y2)

gradopaw:
@chylamal Please fix my comments :)

students/chylak_malgorzata/lesson_02_flow_control/bishop_moves.py
```diff
@@ -0,0 +1,9 @@
+x1 = int(input())
+y1 = int(input())
+x2 = int(input())
+y2 = int(input())
+
+if abs(x1 - x2) == (y1 - y2):
```
gradopaw: should be:
if abs(x1 - x2) == abs(y1 - y2)

PR #290, opened 2018-03-21
adziu/lesson_03_functions

marioosh11:


marioosh11:
@adziu please review

students/mariusz_michalczyk/lesson_03_functions/Calculator.py
```diff
@@ -0,0 +1,59 @@
...
+    print("m - multiply")
+    print("d - divide")
+    print("p - power")
+    print("h,? - help")
+    print("q - QUIT")
+
+
+def get_text(option):
+    if option == "a":
+        return "ADDING"
```
adziu: For consistency, that should be `"ADD"`

students/mariusz_michalczyk/lesson_03_functions/Calculator.py
```diff
@@ -0,0 +1,59 @@
...
+    print(get_text(option))
+    print("Input 1st operand:")
+    var1 = int(input())
+    print("Input 2nd operand:")
+    var2 = int(input())
+    print("Result:")
+    print(get_calculation_result(option, var1, var2))
+
+
+print("Welcome in calculator.")
```
adziu: `main()` would be nice here, so that it's importable.

students/mariusz_michalczyk/lesson_03_functions/Collatz.py
```diff
@@ -0,0 +1,18 @@
...
+    while n > 1:
+        n = get_collatz_nr(n)
+        print(n)
+
+
+try:
+    n = abs(int(input()))
+    run(n)
+except ValueError:
+    print("Please enter integer!!!")
```
adziu: Selbstverständlich!!!

students/mariusz_michalczyk/lesson_03_functions/Negative_exponent.py
```diff
@@ -0,0 +1,7 @@
+def calculate(a, n):
+    return a ** n
```
adziu: I guess what they meant here is to add another `elif` to `exponent.py` and do a division there.
```
def power(a, n):
   if n == 0:
       return 1
   elif n > 0:
       return a * power(a, n - 1)
   else:
       return power(a, n + 1) / a
```

students/urtnowski_daniel/lesson_03_functions/calculator.py
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
```
adziu: You've got some left-overs here, and the module naming should be lowercase, like this one. Also if you create two files with names only differing in case, this might break things for people with Windows.

marioosh11:
Removed not needed calculator.py changes from outside folder. @adziu see if now can be merged

PR #292, opened 2018-03-21
Gradopaw/lesson 02

jarosnat:
@gradopaw


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_02/adding_factorials.py
```diff
@@ -0,0 +1,10 @@
+n = int(input())
+result = 0
+y = 1
+
+for i in range(1, n + 1):
+    for x in range(1, i + 1):
```
gradopaw: Your solution did not pass all of snakify tests.

As it was mentioned in task description:
This problem has a solution with only one loop, so try to discover it. And don't use the math library :)

To do this with one for loop just remove for loop from 6 line
As a result, you will have:

for i in range(1, n + 1):
    y *= i
    result += y

This will repair your solution :)

jarosnat:
@gradopaw

PR #295, opened 2018-03-22
Lesson 2

tomekktomekk:
11 problems are solved.


tomekktomekk:
@adziu Please look on it and let me know what should i fix.

students/kondel_tomasz/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,20 @@
...
+y_2 = int(input())
+
+# there are 8 possible moves from point 1
+
+pos_moves = [x_1 + 1, y_1 + 2], [x_1 + 1, y_1 - 2], [x_1 - 1, y_1 + 2],\
+            [x_1 - 1, y_1 - 2], [x_1 + 2, y_1 + 1], \
+            [x_1 + 2, y_1 - 1], [x_1 - 2, y_1 + 1], [x_1 - 2, y_1 - 1]
+
+res = False
+for point in pos_moves:
```
adziu: You can unpack point here by saying `for p1, p2 in pos_moves:`

students/kondel_tomasz/lesson_02_flow_control/knight_move.py
```diff
@@ -0,0 +1,20 @@
+x_1 = int(input())
+y_1 = int(input())
+x_2 = int(input())
+y_2 = int(input())
+
+# there are 8 possible moves from point 1
+
+pos_moves = [x_1 + 1, y_1 + 2], [x_1 + 1, y_1 - 2], [x_1 - 1, y_1 + 2],\
```
adziu: You don't modify that, so `tuple` would be more appropriate (parentheses instead of brackets)

students/kondel_tomasz/lesson_02_flow_control/the_number_of_zeros.py
```diff
@@ -0,0 +1,13 @@
+N = int(input())
+
+i_list = []
+
+for x in range(N):
+    i_list.append(int(input()))
+
+res = 0
+for y in i_list:
```
adziu: You could also do that with one loop:
```
for _ in range(N):
    y = int(input('Enter a number:))
```

students/kondel_tomasz/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,10 @@
+integer_list = []
+
+input_int = int(input())
+
+while input_int != 0:
+    integer_list.append(input_int)
+    input_int = int(input())
+
+integer_list.sort()
+print(integer_list[len(integer_list) - 2])
```
adziu: There is a shortcut for this: `integer_list[-2]`

students/kondel_tomasz/lesson_02_flow_control/the_second_maximum.py
```diff
@@ -0,0 +1,10 @@
+integer_list = []
+
+input_int = int(input())
+
+while input_int != 0:
+    integer_list.append(input_int)
+    input_int = int(input())
+
+integer_list.sort()
```
adziu: This will work, but you will use O(n) memory and O(n * log(n)) time, which might be a bummer with larger data.

students/kondel_tomasz/lesson_02_flow_control/queen_move.py
```diff
@@ -0,0 +1,31 @@
...
+                 (y_2 - y_1) * (x_1_minus_45_deg - x_1)
+result_90_deg = (x_2 - x_1) * (y_1_90_deg - y_1) - \
+                (y_2 - y_1) * (x_1_90_deg - x_1)
+result_0_deg = (x_2 - x_1) * (y_1_0_deg - y_1) - \
+               (y_2 - y_1) * (x_1_0_deg - x_1)
+
+
+if (result_plus45 == 0) or (result_minus45 == 0) or \
+        (result_90_deg == 0) or (result_0_deg == 0):
+    print('YES')
```
adziu: The same comment as with `bishop_move` applies to this:
`print('YES' if abs(x1 - x2) == abs(y1 - y2) or x1 == x2 or y1 == y2 else 'NO')`

PR #304, opened 2018-03-23
Lesson 03 functions

Julitt:
@tomekDiakowTieto

### DIFF
--- a/students/bedkowska_julita/lesson_03_functions/Exponentiation.py
+++ b/students/bedkowska_julita/lesson_03_functions/Exponentiation.py
@@ -1,0 +1,12 @@
+base = float(input())
+expon = int(input())
+
+
+def exponentiation(a, n):
+    if n == 0:
+        return 1
+    else:
+        return a * exponentiation(a, n - 1)
+
+
+print('{} {}'.format('Result:', exponentiation(base, expon)))
--- a/students/bedkowska_julita/lesson_03_functions/The_length_of_the_segment.py
+++ b/students/bedkowska_julita/lesson_03_functions/The_length_of_the_segment.py
@@ -7,7 +7,7 @@ y2 = float(input('Give y2: '))
 
 
 def distance(x1, y1, x2, y2):
-    return math.sqrt((x2-x1)**2+(y2-y1)**2)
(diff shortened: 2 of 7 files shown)


students/bedkowska_julita/lesson_03_functions/Exponentiation.py
```diff
@@ -0,0 +1,12 @@
+base = float(input())
+expon = int(input())
+
+
+def exponentiation(a, n):
+    if n == 0:
+        return 1
+    else:
+        return a * exponentiation(a, n - 1)
```
tomekDiakowTieto: It looks nice but in python You can:
return 1 if n ==0 else a * exponentiation(a, n-1)

students/bedkowska_julita/lesson_03_functions/The_length_of_the_segment.py
```diff
@@ -0,0 +1,13 @@
+import math
+
+x1 = float(input('Give x1: '))
+y1 = float(input('Give y1: '))
+x2 = float(input('Give x2: '))
+y2 = float(input('Give y2: '))
+
+
+def distance(x1, y1, x2, y2):
+    return math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
```
tomekDiakowTieto: TIP:
>>> 2 ** 0.5
1.4142135623730951
>>> 4 ** 0.5
2.0
>>> 16 ** 0.5
4.0

students/bedkowska_julita/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if option == "p":
+        print("POWER")
+        var_1, var_2 = get_values()
+        print(var_1 ** var_2)
+
+    if option == "h" or option == "?":
+        print("HELP")
+        print(menu())
+
```
tomekDiakowTieto: Ok, but You can build functions like:
def add(a, b):
    return a + b

later You can build dict:
options = {
    'a': add,
    ...
}

later You can type:
options[optio](10, 30);

PR #305, opened 2018-03-24
Added: Lesson 04 - unit testing

rafalskowronski:
@tomekDiakowTieto Lesson 04 ready


students/skowronski_rafal/lesson_04_unit_testing/project_01_collatz/collatz.py
```diff
@@ -0,0 +1,24 @@
+# Project: The Collatz Sequence
+
+
+def collatz(number):
+    if number % 2 == 0:
+        ret_number = number // 2
+    else:
+        ret_number = 3 * number + 1
+
+    return ret_number
```
tomekDiakowTieto: tip:
return number //2 if number % 2 == 0 else 3 * number + 1
but propably Your solution is more readable :)

rafalskowronski: Agree:)

PR #306, opened 2018-03-24
Lesson 05: Add solutions for all problems

mielcszy:
@kuszcjan: done - please review


students/mielczarek_szymon/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,12 @@
+a = [int(i) for i in input().split()]
+min_idx = 0
+max_idx = 0
+
+for i in range(1, len(a)):
```
kuszcjan: good, 
however, in my opinion, it would be better and more readable to avoid iteration and use following:
`min_idx = index(min(a))`
so the full solution would have less lines:
```
a = [int(i) for i in input().split()]

min_idx = a.index(min(a))
max_idx = a.index(max(a))

a[min_idx], a[max_idx] = a[max_idx], a[min_idx]
print(' '.join([str(i) for i in a]))
```

students/mielczarek_szymon/lesson_05_lists/the_bowling_alley.py
```diff
@@ -0,0 +1,8 @@
+num_of_pins, num_of_balls = [int(i) for i in input().split()]
+result = ['I'] * num_of_pins
+for i in range(num_of_balls):
```
kuszcjan: no no no, the `i` value in for iteration is not used. 
so by convention, it should be underscore `_`
please remember that!

students/mielczarek_szymon/lesson_05_lists/swap_the_columns.py
```diff
@@ -0,0 +1,20 @@
+def swap_columns(a, i, j):
```
kuszcjan: +1 for optional task 👍

PR #309, opened 2018-03-25
Lesson 04 - three tasks

karinanartowska:


karinanartowska:
@gradopaw please review :)

students/nartowska_karina/lesson_04_unit_testing/test_for_fibonacci_numbers.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+
+
+def fib(n):
```
gradopaw: Lines from 4 to 21 are not needed - the aim of this task is to test only fib function:) You can just import fib function from the fibonacci_numbers module from lesson_03_functions. To do it:
- mark your folder (nartowska_karina) as src folder (go to PyCharm-> right click on the karina_nartowska folder -> Mark Directory as -> choose source root
- Add empty __init__.py file in the lesson_03_functions to treat this directory as containing packages (have a look at https://docs.python.org/3/tutorial/modules.html#packages)
- now in the test_for_fibonacci_numbers,py you should have the possibility to import function you would like to test (from lesson_03_functions.fibonacci_numbers import fib)

students/nartowska_karina/lesson_04_unit_testing/unit_tests_for_distance.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+from math import sqrt
+
+
+def distance(x1, y1, x2, y2):
```
gradopaw: Look at my comment to test_for_fibonacci_numbers.py. Before you will import distance function you should add a fix to it (function should return value:))

students/nartowska_karina/lesson_04_unit_testing/unit_tests_for_collatz.py
```diff
@@ -0,0 +1,38 @@
+import unittest
+
+
+def collatz(number):
```
gradopaw: Look at my comment to the previous file.

students/nartowska_karina/lesson_04_unit_testing/test_for_fibonacci_numbers.py
gradopaw: the proper name of the empty file you should add "__init__.py". Don't forget to push this file on GitHub as well

karinanartowska:
@gradopaw please review :)

PR #317, opened 2018-03-26
All lesson 3

awulw:
@tomekDiakowTieto

### DIFF
--- a/students/adam_wulw/lesson_03_functions/calculator.py
+++ b/students/adam_wulw/lesson_03_functions/calculator.py
@@ -1,0 +1,97 @@
+def disp_help():
+    """Display function options."""
+    print("Welcome to well organized calculator:")
+    print("a - add")
+    print("s - subtract")
+    print("m - multiply")
+    print("d - divide")
+    print("p - power")
+    print("h,? - help")
+    print("q - QUIT")
+
+
+def math_add():
+    """Reads two int value and print sum of them."""
+    print("ADDING")
+    print("Input 1st operand:")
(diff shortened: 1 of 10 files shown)


students/adam_wulw/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,97 @@
...
+    print("Input 1st operand:")
+    add_var_1 = int(input())
+    print("Input 2nd operand:")
+    add_var_2 = int(input())
+    print("Result:")
+    print(add_var_1 ** add_var_2)
+
+
+disp_help()
+while True:
```
tomekDiakowTieto: You can use sth like: while option != 'q'

students/adam_wulw/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,97 @@
...
+    print("Enter option:")
+
+    option = raw_input()
+
+    if option == "a":
```
tomekDiakowTieto: You can build sth like:
options = {
    'a':  math_add,
    ...
}

and later options[option]()

students/adam_wulw/lesson_03_functions/exponentiation.py
```diff
@@ -0,0 +1,10 @@
+def power(a, n):
+    if n > 0:
+        return a * power(a, n - 1)
+    else:
+        return 1
```
tomekDiakowTieto: TIP: return a * power(a, n -1) if n > 0 else 1

PR #318, opened 2018-03-26
adziu/lesson_04_unit_testing

marioosh11:


students/mariusz_michalczyk/lesson_04_unit_testing/TestCollatz.py
```diff
@@ -0,0 +1,17 @@
...
+import Collatz
+
+
+class TestCollatzFunction(unittest.TestCase):
+    def test_type(self):
+        with self.assertRaises(TypeError):
+            Collatz.get_collatz_nr("aeou")
+
+    def test_expected_collatz_nr(self):
+        self.assertEqual(Collatz.get_collatz_nr(8), 4, "Collatz function returned unexpected value")
```
adziu: That message is too verbose, unittest sucks and will only show you the offending line, but even that is enough to figure out the error. You can put more effort into naming instead, so that when the test fails, you don't even have to read the message. In that exercise it seems a bit strange, but in real code names like `test_validation_succeeds_on_missing_wheel` are a way to go. Try to make it a statement about code for example `test_collatz_returns_4_on_8`.

students/mariusz_michalczyk/lesson_04_unit_testing/TestDistance.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+import The_length_of_the_segment
+
+
+class TestDistanceFunction(unittest.TestCase):
+    def test_type_error_when_string_entered(self):
+        with self.assertRaises(TypeError):
+            The_length_of_the_segment.distance("shit")
+
+    def test_type_error_when_none_entered(self):
```
adziu: This is good.

students/mariusz_michalczyk/lesson_04_unit_testing/TestDistance.py
```diff
@@ -0,0 +1,36 @@
...
+        with self.assertRaises(TypeError):
+            The_length_of_the_segment.distance(None)
+
+    def test_zero_length(self):
```
adziu: This is not so much, it doesn't say what and when it fails, you will have to check the code if the test doesn't pass.

students/mariusz_michalczyk/lesson_04_unit_testing/TestDistance.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertEqual(The_length_of_the_segment.distance(0, 0, 0, 0), 0)
+
+    def test_negative_coordinates(self):
+        self.assertEqual(The_length_of_the_segment.distance(-1, -1, -1, -1), 0)
+        self.assertEqual(The_length_of_the_segment.distance(-1, 0, 0, 0), 1)
+
+    def test_only_vertical_distance(self):
+        with self.assertRaises(TypeError):
+            The_length_of_the_segment.distance(None, 1, None, 2)
```
adziu: Module naming should be all_lower_case

students/mariusz_michalczyk/lesson_04_unit_testing/TestDistance.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_only_horizontal_distance(self):
+        with self.assertRaises(TypeError):
+            The_length_of_the_segment.distance(1, None, 1, None)
+
+    def test_typical_conditions(self):
+        self.assertEqual(The_length_of_the_segment.distance(2, 2, 3, 3), 1.4142135623730951)
```
adziu: This might fail on floats, take a look at this:
https://stackoverflow.com/questions/8929005/unittest-sometimes-fails-because-floating-point-imprecision

PR #319, opened 2018-03-26
Hyska Monika lesson 03

zelazmon:
add new lesson
@adziu


zelazmon:
some technical problems

PR #326, opened 2018-03-27
added lesson 04: unit testing

3tom:
@tomekDiakowTieto , please take a look.

Thanks!
Tomasz


students/medrek_tomasz/lesson_04_unit_testing/fibonacci_unit_tests.py
```diff
@@ -0,0 +1,29 @@
+from fibonacci_numbers import fibonacci
+import pytest
+
+
+def test_none_input_exception():
```
tomekDiakowTieto: You can use:

PR #330, opened 2018-03-27
lesson04

agatastachowska:
@kamwar lesson04 ready to check

### DIFF
--- a/students/stachowska_agata/lesson_04_unit_testing/collatz_sequence_unittest.py
+++ b/students/stachowska_agata/lesson_04_unit_testing/collatz_sequence_unittest.py
@@ -1,0 +1,28 @@
+import unittest
+
+
+def collatz(x):
+    if x % 2 == 0:
+        res = x // 2
+    else:
+        res = 3 * x + 1
+    # print(res)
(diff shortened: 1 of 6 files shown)


students/stachowska_agata/lesson_04_unit_testing/collatz_sequence_unittest.py
```diff
@@ -0,0 +1,28 @@
...
+    if x % 2 == 0:
+        res = x // 2
+    else:
+        res = 3 * x + 1
+    # print(res)
+    return res
+
+
+class TestCollatzFunction(unittest.TestCase):
+
```
kamwar: The empty line is not necessary here

students/stachowska_agata/lesson_04_unit_testing/distance_unittest.py
```diff
@@ -0,0 +1,43 @@
+import unittest
```
kamwar: I know that it was not required to use pytest but try to write unit test using pytest. In modern python project you will most likely use pytest to write unit tests

PR #331, opened 2018-03-27
Lesson 4 - Unit tests

Julitt:
@tomekDiakowTieto


students/bedkowska_julita/lesson_04_unit_testing/test_Fibonacci.py
```diff
@@ -0,0 +1,42 @@
...
+    with pytest.raises(TypeError):
+        fib(None)
+
+
+def test_String_value():
+    with pytest.raises(TypeError):
+        fib('test')
+
+
+def test_zero():
```
tomekDiakowTieto: You can use:

PR #333, opened 2018-03-27
Lesson 04: unit testing

hatan:
@kamwar please check it :)

### DIFF
--- a/students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
+++ b/students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
@@ -12,7 +12,6 @@ def collatz(number):
 
 
 if __name__ == "__main__":
-    print("Please enter a number, positive, integer: ")
-    value = int(input())
+    value = int(input("Please enter a number, positive, integer: "))
     while value != 1:
         value = collatz(value)
--- a/students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
+++ b/students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
@@ -1,0 +1,18 @@
+def collatz(number):
+    if number <= 0:
+        raise ValueError
+    elif number % 2 == 0:
+        half_even = number / 2
+        print(half_even)
+        return half_even
+    elif number % 2 == 1:
+        some_odd = (3 * number) + 1
+        print(some_odd)
+        return some_odd
+
+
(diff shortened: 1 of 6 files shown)


students/kosarzewski_maciej/lesson_03_functions/calculator.py
```diff
@@ -0,0 +1,84 @@
+def welcome():
```
kamwar: Please remove all files related to previous lessons

students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
```diff
@@ -0,0 +1,17 @@
+def collatz(number):
+    if number <= 0:
+        raise ValueError
+    elif number % 2 == 0:
+        half_even = number / 2
+        print(half_even)
+        return half_even
+    elif number % 2 == 1:
```
kamwar: It's better to replace elif with else here. If you really want to keep elif than please remember to handle default case with else
e.g.
else:
    rasie ValueError

students/kosarzewski_maciej/lesson_04_unit_testing/collatz/collatz_sequence.py
```diff
@@ -0,0 +1,17 @@
+def collatz(number):
+    if number <= 0:
+        raise ValueError
+    elif number % 2 == 0:
```
kamwar: elif is not necessary here. Please try to avoid code duplication

def collatz(number):
    if number <= 0:
        raise ValueError
    if number % 2 == 0:
        value = number / 2
    else:
        value = (3 * number) + 1
    print(value)
    return value

students/kosarzewski_maciej/lesson_04_unit_testing/fibonacci/fibonacci_unittest.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_type_error(self):
+        with self.assertRaises(TypeError):
+            fibonacci_number.fib('aoeu')
+
+    def test_type_none(self):
+        with self.assertRaises(TypeError):
+            fibonacci_number.fib(None)
+
+    def test_negative_value(self):
+        with self.assertRaises(RecursionError):
```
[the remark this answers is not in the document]
kamwar: You can ignore this warning now but keep in mind that RecursionError means that a method runs in endless loop which should be generally avoided

hatan:
@kamwar old code removed :)

PR #335, opened 2018-03-27
tasks for lesson_04_unit_testing

miroslawj:

### DIFF
--- a/students/jakielczyk_miroslaw/lesson_04_unit_testing/Collatz_Sequence.py
+++ b/students/jakielczyk_miroslaw/lesson_04_unit_testing/Collatz_Sequence.py
@@ -1,20 +1,19 @@
 # Collatz Sequence
+def collatz(number):
+    if number % 2 == 0:
+        return number // 2
+    elif number % 2 == 1:
+        return 3 * number + 1
+
 
 def main():
-    print ("Enter the number ")
+    print("Enter the number ")
     number = int(input())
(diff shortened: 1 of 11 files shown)


kuszcjan:
First of all, please fix stickler-ci complains.
A more thorough review will be done later.

kuszcjan:
Also, check file names, they are inconsistent and with mixed styles. It's best to use python convection to naming python files.
https://softwareengineering.stackexchange.com/questions/308972/python-file-naming-convention

students/jakielczyk_miroslaw/lesson_04_unit_testing/test_collatz_sequence.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+from collatz_sequence import collatz
+
+
+class CollatzTest(unittest.TestCase):
+
+    def test_1_TypeError(self):
```
kuszcjan: I would still prefer it as `test_1_type_error`

students/jakielczyk_miroslaw/lesson_04_unit_testing/Collatz_Sequence.py
```diff
@@ -0,0 +1,19 @@
+# Collatz Sequence
```
kuszcjan: file names(python modules) should lowercase

PR #337, opened 2018-03-28
create branch adziu/lesson_04_unit_testing

RCanDo:
@adziu

### DIFF
--- a/students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_collatz.py
+++ b/students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_collatz.py
@@ -1,0 +1,25 @@
+import unittest
+
+
+def collatz(number):
+    if number % 2 == 0:
+        return number // 2
+    else:
+        return 3 * number + 1
+
+
+class TestCollatz(unittest.TestCase):
+
+    def test_type(self):
+        with self.assertRaises(TypeError):
+            collatz('aoeu')
+
+    def test_8_4(self):
+        self.assertEqual(collatz(8), 4, "collatz(8) != 4")
+
+    def test_5_16(self):
+        self.assertEqual(collatz(5), 16, "collatz(5) != 16")
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_distance.py
+++ b/students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_distance.py
@@ -1,0 +1,29 @@
+import pytest
+
+
+def distance(x1, y1, x2, y2):
+    return ((x1 - x2) ** 2 + (y1 - y2) ** 2) ** (1 / 2)
+
+
+class TestClass(object):
+
+    def test_none(self):
+        with pytest.raises(TypeError):
+            distance(None, 0, 0, 1)
+
+    def test_value_error(self):
+        with pytest.raises(ValueError):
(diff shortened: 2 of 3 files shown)


students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_collatz.py
```diff
@@ -0,0 +1,25 @@
...
+def collatz(number):
+    if number % 2 == 0:
+        return number // 2
+    else:
+        return 3 * number + 1
+
+
+class TestCollatz(unittest.TestCase):
+
+    def test_type(self):
```
adziu: Try to make a name a statement about the code `test_collatz_raises_TypeError_on_aoeu`, Dvorak bro! This way when the test fails you can jump to code directly.

students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_distance.py
```diff
@@ -0,0 +1,29 @@
...
+            distance(None, 0, 0, 1)
+
+    def test_value_error(self):
+        with pytest.raises(ValueError):
+            distance('aoeu', 0, 0, 1)
+
+    def test_corners(self):
+        assert distance(0, 0, 0, 0) == 0
+        assert distance(-1, -2, -2, -1) > 0
+        assert distance(0, 1, 0, 2) == 1
```
adziu: Floating point arithmetic may fail you here, use something like `math.fabs(result - expected) < epsilon`.

students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_distance.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+
+def distance(x1, y1, x2, y2):
+    return ((x1 - x2) ** 2 + (y1 - y2) ** 2) ** (1 / 2)
+
+
+class TestClass(object):
```
adziu: You don't need that in `pytest`.

students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_collatz.py
```diff
@@ -0,0 +1,25 @@
...
+        with self.assertRaises(TypeError):
+            collatz('aoeu')
+
+    def test_8_4(self):
+        self.assertEqual(collatz(8), 4, "collatz(8) != 4")
+
+    def test_5_16(self):
+        self.assertEqual(collatz(5), 16, "collatz(5) != 16")
```
adziu: A message is a good idea for unittest, pytest should do it automatically.

students/arkadiusz_kasprzyk/lesson_04_unit_testing/test_distance.py
```diff
@@ -0,0 +1,29 @@
...
+        with pytest.raises(TypeError):
+            distance(None, 0, 0, 1)
+
+    def test_value_error(self):
+        with pytest.raises(ValueError):
+            distance('aoeu', 0, 0, 1)
+
+    def test_corners(self):
+        assert distance(0, 0, 0, 0) == 0
+        assert distance(-1, -2, -2, -1) > 0
```
adziu: Not very strict... math.fabs(result) < epsilon would be better. Think about it as something that could save your ass from searching for an error, when you are already using that function. You want to trust your code, so you need to check it hard.

PR #338, opened 2018-03-28
Lesson 3

Graheb:
@kamwar


students/Hebdowska_Grazyna/lesson_03_functions/Exponentiation.py
```diff
@@ -0,0 +1,10 @@
+def power(a, n):
```
kamwar: Please use lower case for the file name

PR #339, opened 2018-03-28
Solutions for exercises in lesson 04

BartoszPi:
@gradopaw please review

Note that in task 2 point ii I checked for TypeError instead of ValueError as checking for ValueError seemed to make no sense.


students/pietrewicz_bartosz/lesson_04_unit_testing/fibonacci_numbers.py
```diff
@@ -0,0 +1,20 @@
+def fib(n):
```
gradopaw: The aim of this lesson is to test some of your solutions from lesson 3. Adding one more time same solutions in this lesson is not needed. To test your fibonacci_numbers module from the previous lesson do following things:

- mark your folder (pietrewicz_bartosz) as src folder (go to PyCharm-> right click on the pietrewicz_bartosz folder -> Mark Directory as -> choose source root

- add empty __init__.py file in the lesson_03_functions to treat this directory as containing packages (have a look at https://docs.python.org/3/tutorial/modules.html#packages) and don't forget to push this file on GitHub as well

- now in the test_fib_numbers.py you should have the possibility to import function you would like to test (from lesson_03_functions.fibonacci_numbers import fib)

students/pietrewicz_bartosz/lesson_04_unit_testing/test_collatz.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from the_collatz_sequence import collatz
+
+
+class TestCollatzFunction(unittest.TestCase):
+    def test_invalid_type(self):
+        self.assertRaises(TypeError, collatz, 'aoeu')
+
+    def test_value_8(self):
+        self.assertEquals(collatz(8), 4)
```
gradopaw: assertEquals is deprecated - use assertEqual instead :) 
Have a look at:
https://docs.python.org/3.6/library/unittest.html#deprecated-aliases

BartoszPi:
@gradopaw please review

PR #340, opened 2018-03-29
Tomek diakow tieto/lesson 04 unit testing

awulw:
@tomekDiakowTieto

### DIFF
--- a/students/adam_wulw/lesson_04_unit_testing/test_fib_num.py
+++ b/students/adam_wulw/lesson_04_unit_testing/test_fib_num.py
@@ -1,0 +1,17 @@
+from fib_num import fib
+
+
+def test_fib_1():
+    assert fib(1) == 1
+
+
+def test_fib_2():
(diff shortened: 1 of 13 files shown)


students/adam_wulw/lesson_04_unit_testing/test_fib_num.py
```diff
@@ -0,0 +1,17 @@
+from fib_num import fib
+
+
+def test_fib_1():
```
tomekDiakowTieto: You can use:

students/adam_wulw/lesson_04_unit_testing/test_pytest_distance.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+from the_length_of_segment import distance
+
+
+def test_type_error():
```
tomekDiakowTieto: You can put everything in TestClass

PR #342, opened 2018-03-29
lesson_03

jarosnat:
@gradopaw

### DIFF
--- a/students/jarosz_natalia/lesson_03/calculator.py
+++ b/students/jarosz_natalia/lesson_03/calculator.py
@@ -55,7 +55,6 @@ while True:
     elif option == "h" or option == "?":
         help_function()
 
-
     else:
         print("Input 1st operand:")
         add_var_1 = int(input())
--- a/students/jarosz_natalia/lesson_03/calculator.py
+++ b/students/jarosz_natalia/lesson_03/calculator.py
@@ -44,42 +44,46 @@ def help_function():
     print("q - QUIT")
 
 
-while True:
-    print("Enter option:")
+def main():
+    while True:
+        print("Enter option:")
+        option = input()
 
-    option = input()
+        if option == "q":
+            print("GOOD BYE")
+            break
+        elif option == "h" or option == "?":
+            help_function()
 
-    if option == "q":
-        print("GOOD BYE")
-        break
-    elif option == "h" or option == "?":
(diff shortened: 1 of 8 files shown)


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_03/calculator.py
```diff
@@ -0,0 +1,89 @@
...
+print("h,? - help")
+print("q - QUIT")
+
+
+def adding(add_var_1, add_var_2):
+    res = add_var_1 + add_var_2
+    return res
+
+
+def substraction(add_var_1, add_var_2):
```
gradopaw: Typo: subtraction

students/jarosz_natalia/lesson_03/calculator.py
```diff
@@ -0,0 +1,89 @@
+print("Welcome to badly organized calculator:")
```
gradopaw: I this line could be moved to the line 48 before ```while``` statement, Than in line 49 you can call ```help()``` and than lines from 2 to 8 could be deleted:)

students/jarosz_natalia/lesson_03/input_validation.py
```diff
@@ -0,0 +1,21 @@
...
+    if arg % 2 == 0:
+        print(arg // 2)
+        return arg // 2
+    else:
+        print(3 * arg + 1)
+        return (3 * arg) + 1
+
+
+def main():
+    print(provide number)
```
gradopaw: This print does not work. Should be ```print('provide number')```

jarosnat:
@gradopaw

students/jarosz_natalia/lesson_03/calculator.py
```diff
@@ -0,0 +1,86 @@
...
+            print("GOOD BYE")
+            break
+        elif option == "h" or option == "?":
+            help_function()
+
+        else:
+            if option != 'a' and option != 's' and option != 'm' and option\
+                    != 'd' and option != 'p':
+                print("Wrong choice, choose another option.")
+                continue
```
gradopaw: Instead of checking twice the value of ```option``` better is add ```else``` in line 81 where you just prin message and continue loop :) This is just tip :)

PR #343, opened 2018-03-29
Lesson 5 solutions

synulewar:
@kuszcjan


students/synowiec_krzysztof/lesson_05_lists/numbers_filter.py
```diff
@@ -0,0 +1,9 @@
+def main():
+    strings = [x for x in input().split()]
+    filter = range(int(input("Provide filter\n")))
```
kuszcjan: `filter` shadows built-in name

students/synowiec_krzysztof/lesson_05_lists/numbers_filter.py
```diff
@@ -0,0 +1,9 @@
...
+    ints = [int(strings[x]) for x in range(len(strings)) if x not in filter]
```
kuszcjan: Seems that this code filters list based on their indexes. That was not the intention of the task. 
> filters numbers within supplied range

I mean that filtering should be based on actual values, not their position in a list.
Check out updated example:
```
list_of_strings = ['2', '0', '8', '3']	
to_filter_range = range(3)	
expected_output = [8, 3]
```

students/synowiec_krzysztof/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,20 @@
+def main():
+    list = [int(x) for x in input().split()]
```
kuszcjan: shadows built-in function list

students/synowiec_krzysztof/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,20 @@
...
+    minIndex = 0
```
kuszcjan: var name should be lowercase

students/synowiec_krzysztof/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,20 @@
...
+    maxIndex = 0
+    for x in range(1, len(list)):
+        if list[x] > list[maxIndex]:
+            maxIndex = x
+            continue
```
kuszcjan: does `continue` do anything here? There is no more code in `for` block after `if` block. So the flow will be exactly the same without it. Then this just add complexity where it is not necessary.
Bad move :)

students/synowiec_krzysztof/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,20 @@
...
+        elif list[x] < list[minIndex]:
+            minIndex = x
+
+    temp_max = list[maxIndex]
+    list[maxIndex] = list[minIndex]
+    list[minIndex] = temp_max
+    for x in range(0, len(list)):
```
kuszcjan: no need for first argument when it's 0. so 
`for x in range(len(list)):` 
does exactly same thing

students/synowiec_krzysztof/lesson_05_lists/swap_min_and_max.py
```diff
@@ -0,0 +1,20 @@
...
+    minIndex = 0
+    maxIndex = 0
+    for x in range(1, len(list)):
+        if list[x] > list[maxIndex]:
+            maxIndex = x
+            continue
+        elif list[x] < list[minIndex]:
+            minIndex = x
+
+    temp_max = list[maxIndex]
```
kuszcjan: There is common python trick to avoid temporary vars in such situations. 
You can do multiple assignment in single line:
` list[max_idx], list[min_idx] = list[min_idx], list[max_idx]`

students/synowiec_krzysztof/lesson_05_lists/swap_the_columns.py
```diff
@@ -0,0 +1,21 @@
+def main():
```
kuszcjan: nice one

students/synowiec_krzysztof/lesson_05_lists/the_bowling_alley.py
```diff
@@ -0,0 +1,22 @@
+def main():
+    numberOfPins, bowls = [int(x) for x in input().split()]
```
kuszcjan: var name should all be lowercase

kuszcjan:
Overall pretty nice code, but please check my comments

PR #346, opened 2018-03-30
Gradopaw/lesson 03 functions solutions

chylamal:


students/chylak_malgorzata/lesson_03_functions/refactor_calculator.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def subtract():
+    """Subtract function docstring"""
+    print("SUBTRACT")
+    print("Input 1st operand:")
+    add_var_1 = int(input())
+    print("Input 2nd operand:")
+    add_var_2 = int(input())
+    print("Result:")
```
gradopaw: In each function, you repeat lines from 15 to 18 so, they could be moved to separate function.

students/chylak_malgorzata/lesson_03_functions/refactor_calculator.py
```diff
@@ -0,0 +1,120 @@
...
+        if option == "?":
+            help_func()
+
+        if option == "q":
+            exit_func()
+
+            break
+
+        else:
+            print("Wrong name")
```
gradopaw: Almost good :) To have code more optimized you could to use: ```if ... elif .. else ...``` blok:
```
if
    ...
elif
    ...
elif
    ...
elif
    ...
else
    ...
```

PR #348, opened 2018-04-01
Lesson01

ApusAlfa:
Lesson 01 part 1 @tomekDiakowTieto


students/marta_herezo/lesson_01/lesson_01/Area_of_right_angled_triangle.py
```diff
@@ -0,0 +1,13 @@
+# enter the length of the base
+print('enter length of the base')
+c = int(input())
+
+# enter the height
+print('enter the height')
+h = int(input())
+
+P = 0.5 * c * h
```
tomekDiakowTieto: Try to name variables properly:
P - should be lowwer
P - should be 'area', nobody knows what 'p' does it mean :)

students/marta_herezo/lesson_01/lesson_01/Area_of_right_angled_triangle.py
```diff
@@ -0,0 +1,13 @@
...
+
+print('c = ' + str(c))
```
tomekDiakowTieto: Tip: try to use format in future format method

print('c = {0}'.format(c))

students/marta_herezo/lesson_01/lesson_02/car_route.py
```diff
@@ -0,0 +1,15 @@
+# N - kilometer/day
+# M - distance in kilometer
+print('Enter a distance of N kilometers per day = ')
+N = int(input())
+
+print('Enter a length of route = ')
+M = int(input())
+
+r = M % N
```
tomekDiakowTieto: lowwwweeer :)

students/marta_herezo/lesson_01/lesson_02/clock_face_1.py
```diff
@@ -0,0 +1,13 @@
+print('Give hours 0 ≤ H < 12:')
+H = int(input())
+
+print('Give minutes 0 ≤ M < 60:')
+M = int(input())
+
+print('Give seconds 0 ≤ S < 60:')
+S = int(input())
+
+sum = H + M / 60 + S / 3600
```
tomekDiakowTieto: sum is built in function, don't override it :)
>>> tab = [1, 2, 3,4]
>>> sum(tab)
10
>>> sum = 10
>>> sum(tab)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'int' object is not callable
>>>

PR #351, opened 2018-04-03
Lesson 4: unit tests and functions from L3.

kkemot:
@adziu,  I've modified and added 3 files from Lesson 3 because these files haven't been merged.

### DIFF
--- a/students/kondel_tomasz/lesson_03_funtions/the_collatz_sequence.py
+++ b/students/kondel_tomasz/lesson_03_funtions/the_collatz_sequence.py
@@ -25,4 +25,3 @@ def main():
 
 if __name__ == '__main__':
     main()
-
--- a/students/kondel_tomasz/lesson_03_funtions/the_collatz_sequence.py
+++ b/students/kondel_tomasz/lesson_03_funtions/the_collatz_sequence.py
@@ -1,0 +1,28 @@
+def collatz(input_string):
+    try:
+        number = int(input_string)
+    except ValueError:
+        print('Please type int value')
+        return None
+
+    if number % 2 == 0:
+        temp_value = number // 2
+        print(temp_value)
+        return temp_value
+    else:
+        temp_value = 3 * number + 1
+        print(temp_value)
+        return temp_value
+
+
+def main():
+    while True:
+        input_string = input()
+        if collatz(input_string) == 1:
+            print("End")
+            break
+
+
+if __name__ == '__main__':
+    main()
+
--- a/students/kondel_tomasz/lesson_04_unit_testing/collatz_test.py
+++ b/students/kondel_tomasz/lesson_04_unit_testing/collatz_test.py
@@ -1,0 +1,20 @@
+import unittest
+import sys
+sys.path.insert(0, '../lesson_03_funtions/')
+from the_collatz_sequence import collatz
+
+
+class MyFirstUnitTest(unittest.TestCase):
+
+    def test_assertRaises(self):
+        self.assertRaises(collatz('aoeu'))
+
+    def test_output_validation_for_8(self):
+        self.assertTrue(collatz(8) == 4)
+
+    def test_output_validation_for_5(self):
+        self.assertTrue(collatz(5) == 16)
+
+
+if __name__ == '__main__':
+    unittest.main()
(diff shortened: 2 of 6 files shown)


kkemot:
@adziu, How to import function from other directory and avoid below sticker error:
"E402 module level import not at top of file" ?

adziu:
There are several ways:
- use $PYTHONPATH variable outside the script,
- add __init__.py files to your exercise directories and run it with `python -m`,
- put the import in the function...
but they are all hacks in that context, so I guess the best way to do it is just `# noqa: E402`

students/kondel_tomasz/lesson_04_unit_testing/collatz_test.py
```diff
@@ -0,0 +1,20 @@
+import unittest
+import sys
+sys.path.insert(0, '../lesson_03_funtions/')
+from the_collatz_sequence import collatz  # noqa: E402
+
+
+class MyFirstUnitTest(unittest.TestCase):
+
+    def test_assertRaises(self):
+        self.assertRaises(collatz('aoeu'))
```
adziu: It's good to specify the exact assertion you expect.

students/kondel_tomasz/lesson_04_unit_testing/collatz_test.py
```diff
@@ -0,0 +1,20 @@
+import unittest
+import sys
+sys.path.insert(0, '../lesson_03_funtions/')
+from the_collatz_sequence import collatz  # noqa: E402
+
+
+class MyFirstUnitTest(unittest.TestCase):
+
+    def test_assertRaises(self):
```
adziu: With naming unit tests try to be as specific as possible, for example `CollatzTest.test_raises_ValueError_on_string`. That way you can go right away to fix the error in the code without looking at the test code.

students/kondel_tomasz/lesson_04_unit_testing/distance_test.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+import sys
+sys.path.insert(0, '../lesson_03_funtions/')
+from the_length_of_the_segment import distance  # noqa: E402
+
+
+def test_crash_with_TypeError():
```
adziu: This is closer to ideal. `test_raises_TypeError_on_None` or `test_collatz_raises_TypeError_on_None`. The name of the function becomes more important as the number of tests grows to few hundreds. I guess `pytest` doesn't print the name of the module, so I include it... but maybe there is a way to turn it on.

students/kondel_tomasz/lesson_04_unit_testing/distance_test.py
```diff
@@ -0,0 +1,38 @@
...
+    with pytest.raises(ValueError):
+        (distance('aoeu', 0, 0, 0))
+
+
+def test_zero_length():
+    assert(distance(0, 0, 0, 0) == 0)
+
+
+def test_negative_coordinates():
+    assert(distance(-5, -5, -3, -5) > 0)
```
adziu: Not very strict... `math.fabs(result) < epsilon` would be better. Think about it as something that could save your ass from searching for an error, when you are already using that function. You want to trust your code, so you need to check it hard.

students/kondel_tomasz/lesson_04_unit_testing/fibonacci_test.py
```diff
@@ -0,0 +1,20 @@
...
+sys.path.insert(0, '../lesson_03_funtions/')
+from fibonacci_numbers import fib  # noqa: E402
+
+
+class MyUnitTest(unittest.TestCase):
+
+    def test_assertRaises(self):
+        self.assertRaises(fib('aoeu'))
+
+    def test_output_validation_for_8(self):
```
adziu: `test_fib_is_correct_for_8`. You are not checking a validation code here. Passing of the test is a statement about the code.

students/kondel_tomasz/lesson_03_funtions/the_collatz_sequence.py
```diff
@@ -0,0 +1,27 @@
+def collatz(input_string):
+    try:
+        number = int(input_string)
```
adziu: So it's my stupid formulation of the exercise that made you take strings here, which is stupid. Sorry for that.

PR #355, opened 2018-04-06
Lekcja 3

kulesmon:
@tomekDiakowTieto

### DIFF
--- a/students/kulesza_monika/Lesson_03.py
+++ b/students/kulesza_monika/Lesson_03.py
@@ -1,0 +1,119 @@
+output = 0
+num1 = ""
+num2 = ""
+choice = ""
+num1 = input("Input first operand:\n")
+num2 = input("Input second operand:\n")
+
+floatnum1 = float(num1)
+floatnum2 = float(num2)
+
+def add(x, y):
+    return x + y
+def subtract(x, y):
+    return x - y
+def multiply(x, y):
+    return x * y
+def divide(x, y):
+    return x / y
+def power (x, y):
+    return x ** y
+
+print("Welcome to badly organized calculator:\n"
+              "a - add\n"
+              "s - subtract\n"
+              "m - multiply\n"
+              "d - divide\n"
+              "p - power\n"
+              "h,? - help\n"
+              "q - QUIT\n")
+
+choice = input("Enter choice your choice: ")
+num1 = int(input("Enter First number:  "))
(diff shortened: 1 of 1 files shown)


students/kulesza_monika/Lesson_03.py
```diff
@@ -0,0 +1,119 @@
...
+              "d - divide\n"
+              "p - power\n"
+              "h,? - help\n"
+              "q - QUIT\n")
+
+choice = input("Enter choice your choice: ")
+num1 = int(input("Enter First number:  "))
+num2 = int(input("Enter Second number:  "))
+
+if choice == 'a':
```
tomekDiakowTieto: This choice should be in 
while choice != 'q'

students/kulesza_monika/Lesson_03.py
```diff
@@ -0,0 +1,119 @@
...
+
+def add(x, y):
+    return x + y
+def subtract(x, y):
+    return x - y
+def multiply(x, y):
+    return x * y
+def divide(x, y):
+    return x / y
+def power (x, y):
```
tomekDiakowTieto: Blank lines between methods

students/kulesza_monika/Lesson_03.py
```diff
@@ -0,0 +1,119 @@
...
+          "q - QUIT\n")
+
+elif choice == 'q':
+    print ("goodbye")
+    break
+
+else:
+    print("Invalid input")
+
+------------------------------------
```
tomekDiakowTieto: If You want to seperate code use
############# instead of -----------------

students/kulesza_monika/Lesson_03.py
```diff
@@ -0,0 +1,119 @@
...
+
+def the_lenght_of_the_segment():
+    def distance(x1, y1, x2, y2):
+        from math import sqrt
```
tomekDiakowTieto: Move this import outside function

students/kulesza_monika/Lesson_03.py
```diff
@@ -0,0 +1,119 @@
...
+    def power(a, n):
+        if n == 0:
+            return 1
+        else:
+            return a * power(a, n - 1)
+    print(power(float(input()), int(input())))
+
+    ------------------------------------
+
+def fibonacci_numbers():
```
tomekDiakowTieto: Try to do it using recursion

PR #362, opened 2018-04-09
Lesson 05 lists

jemiekam:
@kamwar

### DIFF
--- a/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
+++ b/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
@@ -8,7 +8,8 @@ grid = [['.', '.', '.', '.', '.', '.'],
         ['.', 'O', 'O', '.', '.', '.'],
         ['.', '.', '.', '.', '.', '.']]
 
-for first_loop in range(0, len(grid[0])):
-    for second_loop in range(0, len(grid)):
-        print(grid[second_loop][first_loop], end="")
-    print()
+for first_loop in range(len(grid[0])):
+    line = ""
+    for second_loop in range(len(grid)):
+        line += grid[second_loop][first_loop]
+    print(line)
--- a/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
+++ b/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
@@ -12,4 +12,3 @@ for first_loop in range(0, len(grid[0])):
     for second_loop in range(0, len(grid)):
         print(grid[second_loop][first_loop], end="")
     print()
-
--- a/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
+++ b/students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
@@ -1,0 +1,15 @@
+grid = [['.', '.', '.', '.', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['.', 'O', 'O', 'O', 'O', 'O'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for first_loop in range(0, len(grid[0])):
+    for second_loop in range(0, len(grid)):
+        print(grid[second_loop][first_loop], end="")
+    print()
+
--- a/students/jemielity_kamil/lesson_05_lists/comma_code.py
+++ b/students/jemielity_kamil/lesson_05_lists/comma_code.py
@@ -1,0 +1,14 @@
+def comma_code(elements):
+    sentence = ''
+    length = len(elements)
+    for i in range(length):
+        if i is length-1:
+            sentence += 'and ' + elements[i]
+        else:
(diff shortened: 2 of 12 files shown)


students/jemielity_kamil/lesson_05_lists/comma_code.py
```diff
@@ -0,0 +1,14 @@
+def comma_code(elements):
+    sentence = ''
+    length = len(elements)
+    for i in range(length):
+        if i is length - 1:
+            sentence += 'and ' + elements[i]
```
kamwar: For improved readability, it's better to write the code responsible for adding ', ' and then for ' and '

The output is incorrect, it should not includes a comma before ' and' statement:
```
apples, bananas, and tofu
```

My proposal:
```
def comma_code(elements):
    sentence = ''
    length = len(elements)
    for i, element in enumerate(elements):
        if i > 0:
            # Not the first element
            if i < length - 1:
                # Not the last element
                sentence += ', '
            else:
                # Last element
                sentence += ' and '
        sentence += element
    return sentence
```

students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
```diff
@@ -0,0 +1,14 @@
...
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['.', 'O', 'O', 'O', 'O', 'O'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for first_loop in range(0, len(grid[0])):
+    for second_loop in range(0, len(grid)):
+        print(grid[second_loop][first_loop], end="")
```
kamwar: I would avoid using the additional keyword argument 'end' in print method if not necessary. It's also not supported by python < 3

```
for first_loop in range(0, len(grid[0])):
    line = ''
    for second_loop in range(0, len(grid)):
        line += grid[second_loop][first_loop]
    print(line)
```

students/jemielity_kamil/lesson_05_lists/character_picture_grid.py
```diff
@@ -0,0 +1,14 @@
...
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['.', 'O', 'O', 'O', 'O', 'O'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for first_loop in range(0, len(grid[0])):
+    for second_loop in range(0, len(grid)):
```
kamwar: Please use starting argument in range method only if different than 0
range(len(grid[0]))

students/jemielity_kamil/lesson_05_lists/maximum.py
```diff
@@ -0,0 +1,19 @@
...
+for i in range(len(array)):
+    for j in range(len(array[0])):
+        if array[i][j] > maximum:
+            maximum = array[i][j]
+
+for i in range(len(array)):
+    try:
+        print(str(i) + ' ' + str(array[i].index(maximum)))
+        break
+    except ValueError:
```
kamwar: Please avoid handling exception if it's not necessary. It's better to add a condition for checking the index.
Anyway, i and j are already known in the upper code, consider the following:
```
rows_and_columns = [int(i) for i in input('Numbers: ').split()]
m = rows_and_columns[0]  # rows
n = rows_and_columns[1]  # columns

array = [[int(j) for j in input('Numbers in row: ').split()] for k in range(m)]

maximum_i = 0
maximum_j = 0
maximum = array[maximum_i][maximum_j]


for i in range(len(array)):
    for j in range(len(array[0])):
        if array[i][j] > maximum:
            maximum = array[i][j]
            maximum_i = i
            maximum_j = j

print('{} {}'.format(maximum_i, maximum_j))

```

students/jemielity_kamil/lesson_05_lists/swap_the_columns.py
```diff
@@ -0,0 +1,24 @@
+def swap_columns(a, i, j):
+
```
kamwar: Unnecessary new line

students/jemielity_kamil/lesson_05_lists/swap_the_columns.py
```diff
@@ -0,0 +1,24 @@
...
+    for loop in range(len(a)):
+        a[loop][i], a[loop][j] = a[loop][j], a[loop][i]
+
```
kamwar: Unnecessary new line

students/jemielity_kamil/lesson_05_lists/swap_the_columns.py
```diff
@@ -0,0 +1,24 @@
...
+    for i in range(len(a)):
+        for j in range(len(a[0])):
+            a[i][j] = str(a[i][j])
+
```
kamwar: Other than in C, Python doesn't like a new line in front of 'return' or 'for' statements. It's not a mistake but rather a good habit.

students/jemielity_kamil/lesson_05_lists/comma_code.py
kamwar: Sorry, I didn't notice it. You can leave the code as it is

jemiekam:
@kamwar - added fixes

PR #372, opened 2018-04-10
Lesson 6 solutions

synulewar:


synulewar:
@kuszcjan

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/args_sum.py
```diff
@@ -0,0 +1,13 @@
+def main():
+    print(sum_all(1, 2, 3, 4, 5))
+
+
+def sum_all(*args):
+    total = 0
+    for x in args:
+        total += x
+    return total
```
kuszcjan: fine, other way to do same is to use built-in `sum`

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/logger_wrapper.py
```diff
@@ -0,0 +1,18 @@
+from args_inspector import args_inspector
```
kuszcjan: 👍 nice import

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/cubes.py
```diff
@@ -0,0 +1,29 @@
+def main():
+    n, m = [int(x) for x in input().split()]
+    alice_set = prepare_cubes(n)
+    bob_set = prepare_cubes(m)
+
+    intersection_set = sorted(alice_set.intersection(bob_set))
```
kuszcjan: it's totally fine, but just for a record. it's possible to use standard operators here, so this could look like:
`intersection_set = alice_set & bob_set`

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/list_to_fictionary_function_for_fantasy_game_inventory.py
```diff
@@ -0,0 +1,21 @@
...
+def main():
+    inv = {'gold coin': 42, 'rope': 1}
+    dragon_loot = ['gold coin', 'dagger', 'gold coin', 'gold coin', 'ruby']
+    inv = add_to_inventory(inv, dragon_loot)
+    display_inventory(inv)
+
+
+def add_to_inventory(inventory, added_items):
+    for item in added_items:
+        if item not in inventory:
```
kuszcjan: It's fine, but maybe it would be a good idea to use `.get()` here? Something like `inventory.get(item, 0) + 1`
very useful when dealing with python dict
Some https://stackoverflow.com/questions/11041405/why-dict-getkey-instead-of-dictkey

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/frequency_analysis.py
```diff
@@ -0,0 +1,16 @@
+def main():
+    word_dict = {}
+    for x in range(int(input())):
+        for word in input().split():
+            if word not in word_dict:
+                word_dict[word] = 1
+            else:
+                word_dict[word] += 1
+
+    sorted_list = sorted([(-count, word) for word, count in word_dict.items()])
```
kuszcjan: I think I would go for `reverse=True` instead of `-count`

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/the_number_of_distinct_numbers.py
```diff
@@ -0,0 +1,6 @@
+def main():
+    print(len(set(input().split())))
```
kuszcjan: ah, nice one-liner :D

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/uppercase.py
```diff
@@ -0,0 +1,14 @@
+def main():
+    word_list = [x for x in input().split()]
+    for word in word_list:
+        print(capitalize(word), end=" ")
+
+
+def capitalize(word):
+    letters = [x for x in word]
+    letters[0] = letters[0].upper()
+    return ''.join(letters)
```
kuszcjan: I would personally wrote is as:
```
def capitalize(word):
    return word[0].upper() + word[1:]
```

readability in the first place

students/synowiec_krzysztof/lesson_06_dicts_tuples_sets_args_kwargs/access_rights.py
```diff
@@ -0,0 +1,22 @@
...
+    operations_dict = {
+        'read': 'R',
+        'write': 'W',
+        'execute': 'X',
+    }
+
+    file_permissions = {}
+    for x in range(int(input())):
+        file, *operations = input().split()
+        file_permissions[file] = set(operations)
```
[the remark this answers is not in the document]
synulewar: I know that it would work with: "file_permissions[file] = operations" but it seemed to me that there is duplicated permission in one of snakify test. So I decided to use set. I can not find this example know (probably it was never there ;) ) anyway it makes code resistant from permission duplicates.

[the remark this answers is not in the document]
kuszcjan: good and explanation. me like it ;)

PR #373, opened 2018-04-10
Gradopaw/lesson 04

akpoik:
@gradopaw


akpoik:
@gradopaw

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
+import math
```
gradopaw: Don't mix ```unitest``` and ```pytest``` modules. Rewrite this by using only the pytest module. Remember about the proper naming convention.

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
...
+import unittest
```
gradopaw: Line 2 to be deleted

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
...
+import pytest
+from length_of_the_segment import distance
+
+
+class TestDistanceFunction(unittest.TestCase):
```
gradopaw: class TestDistanceFunction(object):

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_input_none(self):
+        with pytest.raises(TypeError):
+            distance(None)
```
gradopaw: I supposed that this test not working as you designed. You have to pass four parameters to ```distance``` function - you passed only one and that's the reason for the raised TypeError exception. Try to run ```distance(None, 1, 2, 3)```

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_input_string(self):
+        with pytest.raises(TypeError):
+            distance('aoeu')
```
gradopaw: As above

students/kulon_aleksander/lesson_04/unit_tests_distance.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_zero_distance(self):
+        self.assertEqual(distance(1, 1, 1, 1), 0, "zero distance")
```
gradopaw: Write this using the pytest module same as the rest of test from this module

students/kulon_aleksander/lesson_04/unit_tests_fibonacci.py
```diff
@@ -0,0 +1,28 @@
+from fibonacci_numbers import fib
```
gradopaw: Same comments as for students/kulon_aleksander/lesson_04/unit_tests_distance.py

akpoik:
@gradopaw

students/kulon_aleksander/lesson_04/test_length_of_the_segment.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_input_string(self):
+        with pytest.raises(TypeError):
+            distance('aoeu', 1, 2, 3)
+
+    def test_zero_distance(self):
+        assert distance(1, 1, 1, 1) == 0
+
+    def test_corner_cases(self):
+        assert distance(-1, -1, -2, -2) == math.sqrt(2)
+        assert distance(0, 1, 0, 2) == 1
```
gradopaw: Tip: Don't make too many assertions per test. Better is to have more tests than one big unit test that is checking few things. A test should only be able to fail for a single reason, namely that the thing it tests doesn’t behave as expected.

PR #374, opened 2018-04-10
Lesson 5

karinanartowska:


students/nartowska_karina/lesson_05_lists/age_calculator.py
```diff
@@ -0,0 +1,41 @@
+def average_age(values):
+    average = 0
+    adult = 0
+    for i in values:
+        if i >= 18:
+            average += i
+            adult += 1
+    if adult > 0:
```
gradopaw: Before if could be one empty line for better readability.

students/nartowska_karina/lesson_05_lists/age_calculator.py
```diff
@@ -0,0 +1,41 @@
...
+        print("The average age of adults: ")
+        return average / adult
+    else:
+        print("The average age of adults: ")
```
gradopaw: Both prints from line 9 and 12 could be placed once in main() function, so this function will only calculate the result.

students/nartowska_karina/lesson_05_lists/age_calculator.py
```diff
@@ -0,0 +1,41 @@
...
+        return 0
+
+
+def count_the_children(values):
+    children = 0
+    for i in values:
+        if i < 18:
+            children += 1
+    print("Count the children: ")
```
gradopaw: Same as previously. You are printing some information string and after then you print result in the main function which is connected with this previously string - it is a bit confusing. Please avoid similar situations in next lessons :)

students/nartowska_karina/lesson_05_lists/chess_board.py
```diff
@@ -0,0 +1,34 @@
...
+    board = []
+    for i in range(rows):
+        row = []
+        for j in range(columns):
+            if i % 2 != j % 2:
+                row.append("*")
+            else:
+                row.append(".")
+        board.append(row)
+    print("Result:")
```
gradopaw: Same as previously, better is to print this in main() where is also printed result.

PR #378, opened 2018-04-11
Add excersises lesson_05_lists

marioosh11:


marioosh11:
@adziu please review

students/mariusz_michalczyk/lesson_05_lists/comma_code.py
```diff
@@ -0,0 +1,9 @@
+spam = ['apples', 'bananas', 'tofu', 'cats']
+
+
+def get_comma_string(my_list):
+    my_list[-1] = "and " + my_list[-1]
```
adziu: I was wondering if this shouldn't require the list to be at least 2 elements long, but they didn't specify it, so let it be.

students/mariusz_michalczyk/lesson_05_lists/numbers_filter.py
```diff
@@ -0,0 +1,8 @@
+strs = ['2', '0', '8', '3']
+
+
+def filter_nrs(strs, range_to_filter):
+    return [i for i in strs[range_to_filter - 1:len(strs)]]
+
+
+print(filter_nrs(strs, 3))
```
adziu: I think Janusz (I guess that was from him, as he answered in the channel) meant something different. He meant to filter _out_ range(3) which is 0, 1 and 2, and hence 0 and 2 are out. There were no specs here and not even a function call in an example, so I don't care... but you would probably get something around "TypeError: can't substract string from a generator".

students/mariusz_michalczyk/lesson_05_lists/maximum.py
```diff
@@ -0,0 +1,22 @@
+def get_array_based_on_user_input():
```
adziu: Why not `array_from_input` (God kills a kitten for every preposition, "get", "generic" and "abstract" :)

PR #380, opened 2018-04-11
Added lesson 05 solutions

swietmon:
@kamwar Please check


students/swietczak_monika/lesson_05_lists/character_picture_grid.py
```diff
@@ -0,0 +1,16 @@
...
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for x in range(6):
+    for y in range(9):
+        if y < 8:
+            print(grid[y][x], end='')
+        else:
+            print(grid[y][x])
```
kamwar: I would avoid using the additional keyword argument 'end' in print method if it's not necessary. It's also not supported by python < 3

Consider the following example:
```
for x in range(len(grid[0])):
    line = ""
    for y in range(len(grid)):
        line += grid[y][x]
    print(line)
```

students/swietczak_monika/lesson_05_lists/greater_than_neighbours.py
```diff
@@ -0,0 +1,10 @@
+lista = input("Enter a list of integer numbers separated by space: ")
+nowa_lista = lista.split()
```
kamwar: Please try to not mix Polish and English words

swietmon:
@kamwar Changes done

PR #383, opened 2018-04-11
Adziu/lesson 05 lists

RCanDo:
@adziu


students/arkadiusz_kasprzyk/lesson_04_unit_testing/.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,7 @@
+{
```
adziu: Add `.*` to `.gitignore` to avoid commiting those things by mistake.

students/arkadiusz_kasprzyk/lesson_05_lists/bowling_alley.py
```diff
@@ -0,0 +1,48 @@
...
+                raise ValueError
+
+        pins = list('I' * N)
+
+        for r in rolls:
+            r[0] -= 1
+            for k in range(*r):
+                pins[k] = '.'
+
+    return(pins)
```
adziu: `''.join(pins)`. Given that I was rather picky about all the `input`s in your code, I should be happy here.

students/arkadiusz_kasprzyk/lesson_05_lists/charachter_picture_grid.py
```diff
@@ -0,0 +1,17 @@
...
+        ['.', 'O', 'O', 'O', 'O', 'O'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for y in range(len(grid[0])):
+    for x in range(len(grid)):
+        print(grid[x][y], end="")
+    print('')
```
adziu: `print(''.join(''.join(zip(*grid))))`? Just kiddin'

students/arkadiusz_kasprzyk/lesson_05_lists/chess_board.py
```diff
@@ -0,0 +1,57 @@
+def rep(x, length=None, times=None):
+    """
+    Replicates x (list or tuple) to the given length or times many.
```
adziu: The idea is to write imperative and use statements as names in unit tests. So: `Replicate a sequence to fill length`. And then in `test_abc_is_replicated_3_times`.

students/arkadiusz_kasprzyk/lesson_05_lists/comma_code.py
```diff
@@ -0,0 +1,30 @@
...
+    s = ''
+    l = len(lst)
+
+    if l > 0:
+        s += str(lst[0])
+        if l > 1:
+            sep = [', '] * (l - 2) + [' and ']
+            for k in range(l - 1):
+                s += (sep[k] + str(lst[k + 1]))
+    return s
```
adziu: I find it very confusing, especially since the naming is too short. Why not:
```
if len(lst) > 1:
  *all_but_last, last = lst
  return ', '.join(all_but_last) + ', and ' + last
else:
  return lst
```

adziu:
Also needs a rebase to master, I just merged your lesson 4 (sorry for delay).

students/arkadiusz_kasprzyk/lesson_05_lists/charachter_picture_grid.py
RCanDo: I tried it and it doesn't work...
The most similar I can imagine is:
     for k in map(lambda t: print(''.join(list(t))) , zip(*grid)): k
what is rather ugly and few levels above...

adziu: I was close... after some testing: `print('\n'.join(map(''.join, zip(*grid))))`

Your version works for me too, the dots are printed.

RCanDo: That's funny... 
Both versions works the same: everything is printed when you run the whole script (Ctrl+Shift+F10) and 3 dots disappear if you run interactively via Python Console (Ctrl+Alt+E). 
So this is a console problem.
Thanks for the tip; I didn't know that `lambda` keyword may be ommited, and that .join() works on map :) (hence on any iterable i guess).

adziu: The only times you need to unpack iterables is debugging and preparing the end result `list()` or `tuple()` works great for the latter.

adziu: Strange. The console might interpret that as continuation line from IPython.

RCanDo: Certainly, but how to change this behaviour? (probably very hard to find it, so it doesn't matter, but it worth knowing that such things happen)

students/arkadiusz_kasprzyk/lesson_05_lists/comma_code.py
RCanDo: OK, but the second line `*all_but_last, last = lst` is tricky if you don't know it's possible to use `*` this way. I can't remember any such example before. Again, thanks for the tip!

I suffer a lot inventing names and unfortunatelly, I'm matematician 😁 so I love one-letter names. Sorry... I understand the problem in larger projects but in few-lines code... Never mind! I'll try to invent better names.

adziu: That's a new thing with Python3, I think it's the clearest form of head and tail from functional languages. I never used it in real code yet and it's not obvious.

students/arkadiusz_kasprzyk/lesson_05_lists/chess_board.py
RCanDo: I don't quite get it... 
* unit tests cannot serve as documentation (they are not part of a file/source code and may be missed), 
* documentation should be maximaly helpful for the user (for newbies too!), while phrases like "Replicate a sequence to fill length" sounds somewhat enigmatic (although very professional!)...
* ... in a sense that, strictly speeking, replicating n times (final length unknown and not an issue) is not the same as filling to the given length.
So imperative is not well defined, and maybe such functions should not be written, but they are useful (I have copied it from basic R). This is matter for a long discussion, again!, how imperatives should be stated, and should they always be atomic (in what sense?). Convienience vs clarity... I don't know! 
Sorry for this! Maybe I just don't understatand what you mean.

adziu: I wasn't picking so much on the formulation (mine is probably worse) as on the grammatical form (fills vs fill). The idea is, don't be so sure what a complicated piece of code is doing, so defensively you should describe it with your wish "replicate string to a given length" instead "replicates strings to a given length".

On the other hand with tests you want to say what the positive result of the test means and try to be specific "joins a strings from an iterable", "joins a strings from a list", "raises an exception on non-iterable" etc.

Unit tests serve as documentation of your non-obvious intentions on edge cases. They are also very useful, when trying to use the code. They show an example of such usage. Also "the documentation" is automatic and will scream at you, if you have misunderstood and changed the intention. I'm always disappointed when reading manual page and `/example` doesn't bring up EXAMPLES section. It means I have to read the boring stuff.

RCanDo: OK, now I got it. It's one of the cases when single 's' has such a deep meaning... 
However I feel that we do not understatnd *documentation* in the same way: for me it is something what serves a final user who has no access to tests (but only to the final product), whereas for you (I guess) it's sth that serves a developer who has access to the whole history of the project (hence to tests too). 
After all, it's very unclear to me how to write docstrings and whom it should serve best. I've read some PEPs, heve seen few conventions in action, PyCharm suggests another, so... ??? Of course there're always sections like *In*, *Out*, *Examples*, *Description* but theirs format and content is rather not precised.

adziu:
My opinion: Docstrings make sense for public APIs and libraries. For
everything else I take the approach from Robert C Martin's Clean Code -- I
try to make code, that doesn't need documentation by making each function
short (under 10 lines) and its name descriptive. If I feel like I failed, I
add a docstring, it's better than nothing.

User documentation and readmes are also useful, but it's way better to
spend time making something easy to use than to document workarounds.

PR #392, opened 2018-04-14
Lesson number 5: Lists

hatan:
@kamwar please check it :)


students/kosarzewski_maciej/lesson_05_lists/character_picture_gird.py
```diff
@@ -0,0 +1,14 @@
...
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['.', 'O', 'O', 'O', 'O', 'O'],
+        ['O', 'O', 'O', 'O', 'O', '.'],
+        ['O', 'O', 'O', 'O', '.', '.'],
+        ['.', 'O', 'O', '.', '.', '.'],
+        ['.', '.', '.', '.', '.', '.']]
+
+for y in range(0, 6):
```
kamwar: Try to avoid adding magic numbers, you might use range(len(grid[0]))
By default range method start from '0' so you can omit the first argument with '0'

students/kosarzewski_maciej/lesson_05_lists/greather_than_neighbours.py
```diff
@@ -0,0 +1,7 @@
+input_data = input().split(' ')
+counter = 0
+for x in range(1, len(input_data) - 1):
+    if input_data[x - 1] < input_data[x] \
+            and input_data[x] > input_data[x + 1]:
```
kamwar: It might be shortened to:
if input_data[x - 1] < input_data[x] > input_data[x + 1]:

Anyway, the current implementation is also correct

PR #399, opened 2018-04-16
Lesson 7 solutions

synulewar:
@kuszcjan

### DIFF
--- a/students/synowiec_krzysztof/lesson_07_string_datetime/to_swap_the_two_words.py
+++ b/students/synowiec_krzysztof/lesson_07_string_datetime/to_swap_the_two_words.py
@@ -1,0 +1,7 @@
+def main():
+    first, second = input().split()
+    print("{0} {1}".format(second, first))
+
+
+if __name__ == '__main__':
+    main()
--- a/students/synowiec_krzysztof/lesson_07_string_datetime/date_calculator.py
+++ b/students/synowiec_krzysztof/lesson_07_string_datetime/date_calculator.py
@@ -1,0 +1,13 @@
+from datetime import datetime, timedelta
+
+
+def main():
+    y, d, h, m = [int(x) for x in input().split()]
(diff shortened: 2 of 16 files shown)


students/synowiec_krzysztof/lesson_07_string_datetime/to_swap_the_two_words.py
```diff
@@ -0,0 +1,7 @@
+def main():
+    first, second = input().split()
+    print("{0} {1}".format(first, second))
```
kuszcjan: This does not swap the words, they are printed exactly the same as input.
@synulewar please fix.

kuszcjan:
just took a quick look. no major problems found so far.

students/synowiec_krzysztof/lesson_07_string_datetime/pythonFile.py
```diff
@@ -0,0 +1,19 @@
+import sys
```
kuszcjan: not sure which task this should solve. maybe was committed by mistake?

synulewar: Yes, that was mistake. It is my script to create empty file with main function. I copy name from snakify or scope and it conver it name to small leters with _. :)

synulewar:
@kuszcjan 
I pushed commit with fix and I can see this change in this pull request but I' m not sure github will refresh and notify.

students/synowiec_krzysztof/lesson_07_string_datetime/pythonFile.py
kuszcjan: hah, understood :)

kuszcjan:
With push force, I don't get a notifications. 
I would prefer additional commit with fixes rather than overwriting single commit. ( and notifications should work too)

PR #401, opened 2018-04-16
Gradopaw/lesson 04

jarosnat:
@gradopaw


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_04/test_the_length_of_the_segment.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+from lesson_03.the_length_of_the_segment import distance
+
+
+class TestDistance(object):
+
+    def test_1(self):
+        with pytest.raises(TypeError):
+            distance(None)
```
gradopaw: I supposed that this test not working as you designed. You have to pass four parameters to distance function - you passed only one and that's the reason for the raised TypeError exception. Try to run distance(None, 1, 2, 3)

jarosnat:
@gradopaw

PR #402, opened 2018-04-17
lesson_02

ApusAlfa:
The full package of lesson_02. @tomekDiakowTieto


students/marta_herezo/lesson_02/for_loop/adding_factorials.py
```diff
@@ -0,0 +1,11 @@
+# Given an integer n, print the sum 1!+2!+3!+...+n!
+print('Enter the number of factorials to add up: ')
+n = int(input())
+factorial = 1
+sum = 0
```
tomekDiakowTieto: sum is built in function, try to use different name

students/marta_herezo/lesson_02/for_loop/adding_factorials.py
```diff
@@ -0,0 +1,11 @@
...
+
+for i in range(1, n + 1):
+    factorial = factorial * i
```
tomekDiakowTieto: You should use *= operator, like +=

students/marta_herezo/lesson_02/for_loop/lost_card.py
```diff
@@ -0,0 +1,17 @@
+# There was a set of cards with numbers from 1 to N.
+# One of the card is now lost.
+# Determine the number on that lost card given
+# the numbers for the remaining cards.
+
+print('Give the numbers of cards: ')
+N = int(input())
```
tomekDiakowTieto: Why this N is upper ?

PR #405, opened 2018-04-17
lesson 3 without cal

dychomon:
@zamber please check lesson 3 without calculator


dychomon:
@zamber all lesson_03_functions with collatz and calculator

PR #413, opened 2018-04-17
Lesson 06

karinanartowska:


karinanartowska:
@gradopaw please review :)

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/uppercase.py
```diff
@@ -0,0 +1,16 @@
+def capitalize(lower_case_word):
+    table_with_letters = [i for i in lower_case_word]
+    first_letter_ascii = ord(table_with_letters[0]) - 32
+    table_with_letters[0] = chr(first_letter_ascii)
+    print(''.join(table_with_letters), end=" ")
+
+
+def main():
+    print("Result:")
```
gradopaw: I think this print could be located before ```for``` loop in line 11. Now, you have situations that "Result" is printed before the user input some value. Please remember such details next time :)

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/uppercase.py
```diff
@@ -0,0 +1,16 @@
+def capitalize(lower_case_word):
+    table_with_letters = [i for i in lower_case_word]
```
gradopaw: The second way to split the word into the list is ```list(lower_case_word)```

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/list_to_dictionary_function_for_fantasy_game_inventory.py
```diff
@@ -0,0 +1,22 @@
+from fantasy_game_inventory import display_inventory
+
+
+def add_to_inventory(inventory, added_items):
+    for i in added_items:
+        if i in inventory:
+            inventory[i] += 1
+        else:
+            inventory[i] = 1
```
gradopaw: Here you could use inventory.setdefault(i, 1) as it was described in the lesson. Your solution is good as well :)

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/the_number_of_distinct_words_in_some_text.py
```diff
@@ -0,0 +1,7 @@
+def main():
```
gradopaw: This code working properly for one line text. According to this task, you should add a possibility to input multi-line text. Please add this.

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/access_rights.py
```diff
@@ -0,0 +1,22 @@
+def main():
+    print("Result:")
+    set_of_operations = {
```
gradopaw: ```set_of_operations``` variable could be global.. Change not needed

students/nartowska_karina/lesson_06_dicts_tuples_sets_args_kwargs/args_sum.py
```diff
@@ -0,0 +1,14 @@
+def sum_all(*args):
+    sum = 0
```
gradopaw: ```sum``` is built-in python function so this name should be avoided (Please have a look at rest of built-in functions: https://docs.python.org/3/library/functions.html). 

This could be done as well using sum function, look at this:
```
def sum_all(*args):
      return sum(args)
```

Of course, your solution is ok

PR #416, opened 2018-04-18
lesson 07 manipulating strings + datetime formatting tasks

jemiekam:
@kamwar

### DIFF
--- a/students/jemielity_kamil/lesson_07_string_datetime/date_calculator.py
+++ b/students/jemielity_kamil/lesson_07_string_datetime/date_calculator.py
@@ -5,5 +5,6 @@ day = int(input('Days to add: '))
 hour = int(input('Hours to add: '))
 minute = int(input('Minutes to add: '))
 
-current_date = datetime.now() + timedelta(weeks=year*52, days=day, hours=hour, minutes=minute)
+current_date = datetime.now() + timedelta(weeks=year * 52, days=day,
+                                          hours=hour, minutes=minute)
 print('Date after change: ' + current_date.strftime("%d-%m-%Y %H:%M:%S"))
--- a/students/jemielity_kamil/lesson_07_string_datetime/date_calculator.py
+++ b/students/jemielity_kamil/lesson_07_string_datetime/date_calculator.py
@@ -1,0 +1,9 @@
+from datetime import datetime, timedelta
+
+year = int(input('Years to add: '))
+day = int(input('Days to add: '))
+hour = int(input('Hours to add: '))
+minute = int(input('Minutes to add: '))
+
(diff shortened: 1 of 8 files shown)


students/jemielity_kamil/lesson_07_string_datetime/date_calculator.py
```diff
@@ -0,0 +1,10 @@
+from datetime import datetime, timedelta
+
+year = int(input('Years to add: '))
+day = int(input('Days to add: '))
+hour = int(input('Hours to add: '))
+minute = int(input('Minutes to add: '))
+
+current_date = datetime.now() + timedelta(weeks=year * 52, days=day,
```
kamwar: Just for your information, please avoid the 'current' word in variable name for something custom and to be placed in the future. It makes more sense for the following  statement 'current_date = datetime.now()'

students/jemielity_kamil/lesson_07_string_datetime/table_printer.py
```diff
@@ -0,0 +1,22 @@
+def print_table(data):
+    col_width = []
+    for i in range(len(data)):
+        maximum = 0
+        for j in range(len(data[0])):
+            if len(data[i][j]) > maximum:
+                maximum = len(data[i][j])
```
kamwar: Try to avoid code duplication, data[i][j] might be assigned to dedicated variable

students/jemielity_kamil/lesson_07_string_datetime/table_printer.py
```diff
@@ -0,0 +1,22 @@
+def print_table(data):
+    col_width = []
+    for i in range(len(data)):
+        maximum = 0
+        for j in range(len(data[0])):
```
kamwar: More pythonic way is to go over items in the list instead of iterating over elements in the list

```
    for row in data:
        maximum = 0
        for item in row:
            if len(item) > maximum:
                maximum = len(item)
        col_width.append(maximum)
```

Or by using list comprehension:
```
max_lengths = [max(len(item) for item in column) for column in data]
```
Even shorten solution with map:
```
max_lengths = [max(map(len, column)) for column in data]
```

students/jemielity_kamil/lesson_07_string_datetime/to_swap_the_two_words.py
```diff
@@ -0,0 +1,4 @@
+string = input('String: ')
+first_word = string[:string.find(' ')]
+second_word = string[string.find(' ') + 1:]
+print(second_word + " " + first_word)
```
kamwar: Please try to use format() in the next lessons.

Just for your information, instead of 'find()' you might also use 'split()'
```
first_word = string.split()[0]
next_words = " ".join(string.split()[1:])
print("{} {}".format(next_words, first_word))
```

PR #420, opened 2018-04-20
added lesson_05

agatastachowska:
@kamwar  ready to check


students/stachowska_agata/lesson_05_lists/greater_than_neighbours.py
```diff
@@ -0,0 +1,8 @@
+n = [int(i) for i in input().split()]
+quantity = 0
+
+for i in range(1, len(n) - 1):
+    if (n[i] > n[i - 1]) and (n[i] > n[i + 1]):
```
kamwar: It might be shortened to:
if n[i + 1] > n[i] > n[i - 1]):
Anyway, the current implementation is also correct

PR #424, opened 2018-04-21
Lesson 06 initial commit

GrzegorzBajorski:
Signed-off-by: Grzegorz Bajorski <Grzegorz.Bajorski@tieto.com>


students/grzegorz_bajorski/lesson_06/access_rights.py
```diff
@@ -0,0 +1,25 @@
...
+
+n = int(input())
+files = {}
+
+for i in range(n):
+    names, *access = input().split()
+    files[names] = set(access)
+
+m = int(input())
+for i in range(m):
```
gradopaw: ```i``` is redeclared and not used. Try to avoid such situations :)

students/grzegorz_bajorski/lesson_06/cities.py
```diff
@@ -0,0 +1,13 @@
+n = int(input())
+cify_of_country = {}
+
+for i in range(n):
+    country, *cities = input().split()
+    cify_of_country[country] = set(cities)
+
+m = int(input())
+for i in range(m):
```
gradopaw: ```i``` is redeclared and not used. Try to avoid such situations :)

students/grzegorz_bajorski/lesson_06/cities.py
```diff
@@ -0,0 +1,13 @@
+n = int(input())
+cify_of_country = {}
```
gradopaw: Typo ```city_of_country```

students/grzegorz_bajorski/lesson_06/sum_all.py
```diff
@@ -0,0 +1,8 @@
+def sum_all(*args):
+    sum = 0
```
gradopaw: Try to avoid using built-in names: https://docs.python.org/3/library/functions.html

Btw built-in function ```sum()``` can be used here as an alternative solution:
```
def sum_all(*args):
      return sum(args)
```

students/grzegorz_bajorski/lesson_06/instpect_args.py
```diff
@@ -0,0 +1,8 @@
+def inspect_args(*args, **kwargs):
```
gradopaw: Typo in the file name.

students/grzegorz_bajorski/lesson_06/logger_wraper.py
```diff
@@ -0,0 +1,8 @@
+def inspect_args(*args, **kwargs):
```
gradopaw: This solution is not compliance with the task. You should write a function, for instance,```logger_wrapper``` which will call ```inspect_args``` and then call another function (foo) that will be passed as a first argument to ```logeer_wrapper()```

```
def logger_wrapper(some_foo, *args, **kwargs ):
      inspect_args(*args, **kwargs)
      some_foo(*args, **kwargs)
```

PR #427, opened 2018-04-22
lesson_05

akpoik:

### DIFF
--- a/students/kulon_aleksander/lesson_05/age_calculator.py
+++ b/students/kulon_aleksander/lesson_05/age_calculator.py
@@ -1,0 +1,17 @@
+def adult(one):
+    return one >= 18
+
+
+def main():
+    people = [int(i) for i in input().split()]
+    adults = [i for i in people if adult(i)]
+    kids = [i for i in people if not adult(i)]
+
+    avg_adults_age = sum(adults) / len(adults)
+    count_kids = len(kids)
+    print("avg_adults_age: %f" % avg_adults_age)
+    print("count_kids: %d" % count_kids)
+
+
+if __name__ == "__main__":
+    main()
--- a/students/kulon_aleksander/lesson_05/age_calculator.py
+++ b/students/kulon_aleksander/lesson_05/age_calculator.py
@@ -3,14 +3,17 @@ def adult(one):
 
 
 def main():
-    people = [int(i) for i in input().split()]
-    adults = [i for i in people if adult(i)]
-    kids = [i for i in people if not adult(i)]
+    try:
+        people = [int(i) for i in input().split()]
+        adults = [i for i in people if adult(i)]
+        kids = [i for i in people if not adult(i)]
 
-    avg_adults_age = sum(adults) / len(adults)
-    count_kids = len(kids)
-    print("avg_adults_age: %f" % avg_adults_age)
-    print("count_kids: %d" % count_kids)
+        avg_adults_age = sum(adults) / len(adults)
+        count_kids = len(kids)
+        print("avg_adults_age: %f" % avg_adults_age)
+        print("count_kids: %d" % count_kids)
+    except ZeroDivisionError:
+        print("Ooops, list cannot be empty!")
 
 
 if __name__ == "__main__":
--- a/students/kulon_aleksander/lesson_05/age_calculator.py
+++ b/students/kulon_aleksander/lesson_05/age_calculator.py
@@ -3,17 +3,18 @@ def adult(one):
 
 
 def main():
-    try:
         people = [int(i) for i in input().split()]
         adults = [i for i in people if adult(i)]
         kids = [i for i in people if not adult(i)]
 
-        avg_adults_age = sum(adults) / len(adults)
+        try:
+            avg_adults_age = sum(adults) / len(adults)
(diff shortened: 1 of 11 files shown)


akpoik:
@gradopaw

students/kulon_aleksander/lesson_05/age_calculator.py
```diff
@@ -0,0 +1,17 @@
+def adult(one):
+    return one >= 18
+
+
+def main():
+    people = [int(i) for i in input().split()]
+    adults = [i for i in people if adult(i)]
+    kids = [i for i in people if not adult(i)]
+
+    avg_adults_age = sum(adults) / len(adults)
```
gradopaw: What in a case when user will not input any adults?

students/kulon_aleksander/lesson_05/chess_board.py
```diff
@@ -0,0 +1,10 @@
+n, m = input().split()
+tab = []
+for i in range(int(n)):
+    tab.append(['* '] * int(m))
+    for j in range(int(m)):
+        if i % 2 == 0 and j % 2 == 0:
+            tab[i][j] = '. '
```
gradopaw: Instead of assigning ```tab[i][j] = '. ' ``` you can omit space at the end so it will be: ```tab[i][j] = '.' ``` and you can join values using space (line 10): ```print(' '.join(tab[i])) ```

students/kulon_aleksander/lesson_05/greater_than_neighbours.py
```diff
@@ -0,0 +1,16 @@
+def func(arg):
+    result = ''
+    return result
+
+
+def main():
+    func(1)
+    a = "-9 29 -100 64 26 73 -96 28 -92 11 -14 -86 -54 -67".split()
```
gradopaw: I suppose that this line is for testing purposes :) To be removed :)

students/kulon_aleksander/lesson_05/greater_than_neighbours.py
```diff
@@ -0,0 +1,16 @@
+def func(arg):
+    result = ''
+    return result
+
+
+def main():
+    func(1)
```
gradopaw: What is a purpose of this function?

students/kulon_aleksander/lesson_05/greater_than_neighbours.py
```diff
@@ -0,0 +1,16 @@
...
+    a = "-9 29 -100 64 26 73 -96 28 -92 11 -14 -86 -54 -67".split()
+    a = input().split()
+    for i in range(len(a)):
```
gradopaw: In this task you should print the quantity of elements that are greater than both of their neighbors not each element. Please change it.

students/kulon_aleksander/lesson_05/maximum.py
```diff
@@ -0,0 +1,19 @@
+m, n = [int(s) for s in input().split()]
+max_i, max_j = m + 1, n + 1
+tab = []
+
+for i in range(m):
+    tab.append(input().split())
+    if i == 0:
+        max = int(tab[0][0])
```
gradopaw: Try to avoid using built-in functions names (https://docs.python.org/3/library/functions.html)

students/kulon_aleksander/lesson_05/swap_min_and_max.py
```diff
@@ -0,0 +1,21 @@
+a = [int(i) for i in input().split()]
+max = 0
+min = 0
+
+for i in range(len(a)):
+    if a[i] > max:
+        max = a[i]
+
+min = max
```
gradopaw: As I mentioned before please don't use built-in functions names.

students/kulon_aleksander/lesson_05/swap_min_and_max.py
```diff
@@ -0,0 +1,21 @@
...
+    if a[i] < min:
+        min = a[i]
+
+index_min = a.index(min)
+index_max = a.index(max)
+
+a[index_min] = max
+a[index_max] = min
+
+print(''.join(a))
```
gradopaw: You can't join integer values, you have to cast here "a" variable to a string so it will be:
```print(''.join(str(a))```

File: students/kulon_aleksander/lesson_05/greater_than_neighbours.py [hunk 9]
```diff
@@ -0,0 +1,15 @@
+def func():
```
gradopaw: Have a look at the code below. We should omit first and last element. Next step is to check if neighbours from both sides are smaller, if yes increase quantity.

```
def func():
    quantity = 0
    a = input().split()
    for i in range(1, len(a) -1):
        if int(a[i-1]) < int(a[i]) > int(a[i+1]):
            quantity += 1
    print(quantity)


def main():
    func()


if __name__ == '__main__':
    main()
```

akpoik: mi sie pomylilo i zle zadanie ze sneakify wklejalem, wklejalem "greater than previous" zamiast "neightbours" :D

akpoik:
@gradopaw

PR #428, opened 2018-04-22
Solve all problems from lesson 05

michalmitek:
@gradopaw Lesson 05 solutions


students/mitek_michal/lesson_05_lists/diagonal_parallel_to_the_main.py
```diff
@@ -0,0 +1,24 @@
+
+def diagonal_parallel_to_the_main():
+
+    n = int(input())
+    array = [[0 for i in range(n)] for j in range(n)]
```
gradopaw: In this line,```i``` and ```j``` are not used. In such cases when you would like to ignore some value use ```_```. Please have a look at https://hackernoon.com/understanding-the-underscore-of-python-309d1a029edc

PR #440, opened 2018-04-25
Results of lesson 05

ssaleta:
@gradopaw After long time I'm back :)


students/saleta_sebastian/lesson_05_lists/comma_code.py
```diff
@@ -0,0 +1,22 @@
+def comma(user_list):
+    if user_list == []:
```
gradopaw: Could be simplified: ```if not user_list```

students/saleta_sebastian/lesson_05_lists/greather_than_neighbours.py
```diff
@@ -0,0 +1,20 @@
+def how_many_is_greather(list_of_numbers):
```
gradopaw: Typo - should be greater (same in the file name) 😃

students/saleta_sebastian/lesson_05_lists/greather_than_neighbours.py
```diff
@@ -0,0 +1,20 @@
...
+    greather_index = 0
+
+    for i in range(1, len(list_of_numbers)):
```
gradopaw: Is it working properly? You should skip also the last element from the list.

students/saleta_sebastian/lesson_05_lists/greather_than_neighbours.py
```diff
@@ -0,0 +1,20 @@
...
+        if list_of_numbers[i - 1] < list_of_numbers[i] > \
+                list_of_numbers[i + 1]:
+            greather_index += 1
+    print(greather_index)
+
+
+def main():
+    a = [int(x) for x in input().split()]
+    print(a)
```
gradopaw: lines from 13  to 15 not needed please remove

students/saleta_sebastian/lesson_05_lists/greather_than_neighbours.py
```diff
@@ -0,0 +1,20 @@
...
+    sample_list = [-9, 29, -100, 64, 26, 73, -96, 28, -92, 11, -14, -86, -54,
+                   -67]
+    how_many_is_greather(sample_list)
```
gradopaw: change sample_lst to a

students/saleta_sebastian/lesson_05_lists/maximum.py
```diff
@@ -0,0 +1,23 @@
+def maximum(row, col, values):
+    maximum = values[0][0]
```
gradopaw: Here you shadow maximum - function name. Please avoid such situations

students/saleta_sebastian/lesson_05_lists/numbers_filter.py
```diff
@@ -0,0 +1,18 @@
+def filter(list, filter_range):
```
gradopaw: filter and list are built-in functions. Try to avoid such names as: https://docs.python.org/3/library/functions.html

students/saleta_sebastian/lesson_05_lists/numbers_filter.py
```diff
@@ -0,0 +1,18 @@
...
+    filter_list = []
+
+    for x in range(len(list)):
+        item = int(list[x])
+        if item not in filter_range:
+            filter_list.append(item)
+    return (filter_list)
```
gradopaw: Redundant parentheses

students/saleta_sebastian/lesson_05_lists/character_picture_grid.py
```diff
@@ -0,0 +1,26 @@
+def get_picture_grid(raw_grid):
```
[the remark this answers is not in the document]
gradopaw: I'm asking because I run your solution with Python 2.7 where it works and when I run with Python 3.5 they do n't work. The problem is that the value returned by built-in zip function is a bit different in both versions 😃

ssaleta:
@gradopaw  ok I think it's correct

PR #445, opened 2018-04-25
tasks for lesson_07_string_datetime

miroslawj:


students/jakielczyk_miroslaw/lesson_07_string_datetime/the_number_of_words.py
```diff
@@ -0,0 +1,7 @@
+def main():
+    string = input()
+    print((string.count(' ') + 1))
```
kuszcjan: little sketchy, may not always work.

PR #446, opened 2018-04-26
'lesson_03'

MaciekHalas:
@tomekDiakowTieto

### DIFF
--- a/lesson_03-6.py
+++ b/lesson_03-6.py
@@ -1,0 +1,18 @@
+import math
+from decimal import Decimal
+
+
+def distance(x1, y1, x2, y2):
+    return math.sqrt(math.pow(x1 - x2, 2) + math.pow(y1 - y2, 2))
+
+
+def program():
+    x1 = Decimal(input())
+    y1 = Decimal(input())
+    x2 = Decimal(input())
+    y2 = Decimal(input())
+    print(distance(x1, y1, x2, y2))
+
+
+if __name__ == '__main__':
+    program()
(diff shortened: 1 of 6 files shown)


lesson_03-6.py
```diff
@@ -0,0 +1,18 @@
+import math
+from decimal import Decimal
+
+
+def distance(x1, y1, x2, y2):
+    return math.sqrt(math.pow(x1 - x2, 2) + math.pow(y1 - y2, 2))
```
tomekDiakowTieto: What about number ** 0.5 :) ?
>>> 16**0.5
4.0
>>>

students/halas_maciek/lesson_01/lesson_03-2.py
```diff
@@ -0,0 +1,22 @@
+def collatz(number):
+    collatzed_number = 0
+    if number % 2 == 0:
```
tomekDiakowTieto: You can use sthi like that:
return number // 2 if number % 2 == 0 else 3 * number + 1

It's only tip because Your solution is propably more readable

students/halas_maciek/lesson_01/lesson_03_functions.py
```diff
@@ -0,0 +1,87 @@
...
+def start_calculator():
+    print("Welcome to badly organized calculator:")
+    show_help()
+    option = None
+
+    while option != "q":
+        print("Enter option:")
+        option = input()
+
+        if option == "a":
```
tomekDiakowTieto: You can use sth like:
options = {
   's': perform_substraction,
   'm': perform_multiplication
   ...
}

and later 
options[option]()

PR #464, opened 2018-05-04
added lesson 08: regular expressions

3tom:
Dear @tomekDiakowTieto 

Please take a look,

Thanks,
Tomasz

### DIFF
--- a/students/medrek_tomasz/lesson_08_regular_expressions/email_validator.py
+++ b/students/medrek_tomasz/lesson_08_regular_expressions/email_validator.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+
+
+import re
+
+
+def email_validator(val_string):
+    multiple_dots_regex = re.compile(r'\.\.')
+
+    last_at = val_string.rfind('@')
+    # Address must contain at least one at sign
+    if last_at == -1:
+        return False
+
+    local_part = val_string[:last_at]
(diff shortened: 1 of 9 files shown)


students/medrek_tomasz/lesson_08_regular_expressions/email_validator.py
```diff
@@ -0,0 +1,61 @@
...
+        return False
+
+    domain_regex = re.compile(r'''(
+        ([a-zA-Z0-9][a-zA-Z0-9.-]+\.)+[a-zA-Z0-9]{2,4}
+    )''', re.VERBOSE)
+    return True if domain_regex.match(domain_part) else False
+
+
+def main():
+    print(email_validator('strange@example.com'))
```
tomekDiakowTieto: We should test non valid email too :)

students/medrek_tomasz/lesson_08_regular_expressions/phone_number_validator.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+
+
+import re
+
+
+def phone_number_validator(validate_string):
+    return True if re.match(r"^(\+\d{2}[- ]?)?(\d{3}[- ]?){2}\d{3}$",
```
tomekDiakowTieto: This regexp is long so we can move it to the variable :)

PR #475, opened 2018-05-07
hyska_monika_add lesson 06

zelazmon:
@adziu


students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/Cubes.py
```diff
@@ -0,0 +1,52 @@
...
+
+# Number of cubes repeated with Alice and Bob
+# Cubes colors repeated with Alice and Bob
+repeated_cubes = list(set(alice_cubes) & set(bob_cubes))
+print("Number repeated cubes colors: ", len(repeated_cubes))
+print("Repeated cubes:")
+print(' '.join(map(str, repeated_cubes)))
+
+# Number of Alice cubes not repeated (Bob hasn't the same colors)
+# Alice cubes colors not repeated (Bob hasn't the same colors)
```
adziu: English: hasn't => doesn't have

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/Cubes.py
```diff
@@ -0,0 +1,52 @@
+"""
+Program use "set", print results for:
+intersection (A & B)
+relative complement of B in A (A - B)
+relative complement of A in B (B - A)
```
adziu: The input is not conforming to the exercise, but I think you get the point of `set()` usage and if necessary you would implement it. So I just let it be.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/FrequencyAnalysis.py
```diff
@@ -0,0 +1,22 @@
...
+def display_sorted_words(words_with_counters):
+    words_with_counters = sorted(words_with_counters,
+                                 key=lambda x: (-x[0], x[1]))
+    print(words_with_counters)
+    for elem in words_with_counters:
+        print(elem[1])
+
+
+# display_sorted_words(list_to_tuples_with_count(file_to_list("text.txt")))
+display_sorted_words(list_to_tuples_with_count(file_to_list("text2.txt")))
```
adziu: Again, intermediate steps and variable assignments would make the code clearer, especially since those are not really functions, but rather procedures that stop for input.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/FrequencyAnalysis.py
```diff
@@ -0,0 +1,22 @@
+# Program display words appearing form the least times to the least
+from functions_lesson06 import file_to_list
+
+
+def list_to_tuples_with_count(word_list):
+    words_with_counters = []
+    for elem in word_list:
```
adziu: You could have put `set()` around `word_list` here.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/FrequencyAnalysis.py
```diff
@@ -0,0 +1,22 @@
...
+        words_with_counters.append((word_list.count(elem), elem))
+        words_with_counters_without_duplication = set(words_with_counters)
```
adziu: English without_duplication => unique

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/GuessTheNumber.py
```diff
@@ -0,0 +1,29 @@
...
+        if numbers == "help":
+            asked_numbers = {key: val for key, val in asked_numbers.items()
+                             if val != 0}
+            break
+        else:
+            numbers = numbers.split(' ')
+            numbers = map(int, numbers)
+            answer = (input("yes/no: "))
+            if answer == "yes":
+                for elem in numbers:
```
adziu: Now this is a place, where `set()` would be great. Something like this (I'm reading the HELP command in the wrong place, but you get the idea):
```
valid = set(range(1,n))
row = set(map(int, input().split(' '))
command = input()
if command == 'YES':
  valid &= row
elif command == 'NO':
  valid -= row
elif command == 'HELP':
   print(valid)
else:
  raise ValueError('Bad command')
```
The clue that you are using the wrong tool is that you have to do many `for` loops.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/Uppercase.py
```diff
@@ -0,0 +1,16 @@
+# Function capitalize(lower_case_word) that takes the lower case word
+# and returns the word with the first letter capitalized
+
+
+def capitalize(lower_case_word):
+    lst = [word[0].upper() + word[1:] for word in lower_case_word.split()]
+    lower_case_word = " ".join(lst)
```
adziu: Very confusing that you are using the same variable for the result.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/ReverseTheSequence.py
```diff
@@ -0,0 +1,17 @@
...
+    if n == 0:
+        print(n)
+        return n
+    else:
+        int_revers()
+        print(n)
+        return n
+
+
+int_revers()
```
adziu: That's a crazy idea... There is a function `reversed`, by the way.

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/functions_lesson06.py
```diff
@@ -44,30 +43,36 @@ def list_elements_str():
...
-        values_elem = split_elements
-        inventory.setdefault(key_elem, values_elem)
+        key, *values = split_elements
+        # -------
+        # doesn't work, when put the same key
+        # inventory.setdefault(key_elem, values_elem)
+        # data wouldn't be save
+        # -------
+        # when put the same key, data are overwritten
+        inventory.update({key: values})
```
adziu: Again, `inventory[key] = values` will do the same thing. `update()` is meant to be used like `old_version_of_the_dict.update(new_entries_which_may_overwrite_some_old_ones)`

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/functions_lesson06.py
```diff
@@ -76,3 +81,13 @@ def file_to_list(file_name):
     my_file = open(file_name, "r")
     words_list = my_file.read().split()
     return words_list
+
+
+# function change dictionary {'Poland' : ['Warsaw', 'Krakow']}
+# to dictionary  {'Warsaw': 'Poland, 'Krakow': 'Poland'}
+def change_dict_without_list(inventory):
```
adziu: English: "change" is not very specific, it may mean almost everything. What you are making here could be called "index" or a "reverse mapping". For me to skip reading the content, I would have to see something like:
```
def city2country(country2cities):
def city2country_mapping(country2city_list):
def get_city2country_mapping(country2cities_mapping):
```

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/ArgsInspector.py
```diff
@@ -0,0 +1,13 @@
...
+
+def inspect_args(*args, **kwargs):
+    print('Arguments:', ', '.join(map(str, args)))
+    print('Keyword :')
+    for k, v in kwargs.items():
+        print(k, "=", v)
+
+
+my_dict = {'o': '8', 'p': 9}
+inspect_args(5, 7, 8, **my_dict, n=9)
```
adziu: That will crash with `InvalidSyntax`, `**keyword_args` must go after all named arguments: `inspect_args(5, 7, 8,  n=9, **my_dict)`

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/ListToDictionaryFunctionForFantasyGameInventory.py
```diff
@@ -0,0 +1,17 @@
+# function to add items to dictionary
+from FantasyGameInventory import display_inventory
+
+
+def add_to_inventory(inventory, added_items):
+    for item in added_items:
+        if item in inventory:
```
adziu: This is the place you could use `setdefault` instead of four lines:
`inventory.setdefault(item, 0) += 1`

adziu: Sorry, it turns out it doesn't work... you would need two lines:
```
inventory.setdefault(item, 0)
inventory[item] += 1

adziu: And another way would be: `inventory[item] = inventory.get(item, 1)`

students/hyska_monika/lesson_06_dicts_tuples_sets_args_kwargs/ArgsInspector.py
adziu: We discussed this and this was changed in Python3 and is now accepted by the compiler.

PR #477, opened 2018-05-07
hyska_lesson_07

zelazmon:
@adziu


students/hyska_monika/lesson_07_manipulating_strings_datetime_formatting/DateCalculator.py
```diff
@@ -0,0 +1,29 @@
...
+# and displays the result in human readable format.
+import datetime
+
+
+def add_to_date(date, years, months, days, hours, minutes):
+    new_date = date + datetime.timedelta(days=days,
+                                         hours=hours,
+                                         minutes=minutes)
+    years += (months + new_date.month) // 12
+    months = (months + new_date.month) % 12
```
adziu: Good that you took care of those edge conditions.

students/hyska_monika/lesson_07_manipulating_strings_datetime_formatting/DatePrinter.py
```diff
@@ -0,0 +1,28 @@
...
+print('Second: %d' % now.second)
+print('Microsecond: %d' % now.microsecond)
+
+print("\nCurrent date and time using strftime:")
+print(now.strftime("%Y-%m-%d %H:%M"))
+print(now.strftime("%d.%m.%Y %H:%M"))
+print(now.strftime("%d %B %Y %H:%M"))   # month by word
+
+print("\nCurrent date and time using isoformat:")
+print(now.isoformat())
```
adziu: :+1:

students/hyska_monika/lesson_07_manipulating_strings_datetime_formatting/DeltaTimeCalculator.py
```diff
@@ -0,0 +1,17 @@
...
+
+def different_in_days(date, date2):
+    delta = date2 - date
+    print('Different in days: ', delta.days)
+    return delta.days
+
+
+now = datetime.datetime.now()
+feature_date = datetime.datetime(2020, 10, 6, 13, 22)
+print('Actual date and time: ', now.strftime("%Y-%m-%d %H:%M"))
```
adziu: English: "actual" is a false friend, what you mean is "current"

students/hyska_monika/lesson_07_manipulating_strings_datetime_formatting/TablePrinter.py
```diff
@@ -0,0 +1,21 @@
+# function print table with each column right-justified
+
+from functions_lesson07 import print_table, find_max_sting
```
adziu: If you are using those functions only here, then it's kind of strange to have a separate module for them.

students/hyska_monika/lesson_07_manipulating_strings_datetime_formatting/functions_lesson07.py
```diff
@@ -0,0 +1,30 @@
...
+    # longest word(max), and return strings length
+    max_table = len(max(sum(my_dict, []), key=len))
+    return max_table
+
+
+# function return length for the longest string in array
+def find_max_sting2(my_dict):
+    max_str = len(my_dict[0][0])
+    for row in my_dict:
+        max_row = len(max(row, key=len))
```
adziu: This is a maximum length of a cell in a row and in the end the function still returns one number instead of a list to use with the `rjust`. This is still not what is meant. You need a maximum cell length in each column, and then in each row for each cell you use max_column_length for this particular column.
```
max_column_length = [max([len(word) for word in column]) for column in zip(*table_data)]
...
        print(' '.join([cell.rjust(max_column_length[i]) for i, cell in enumerate(row)]))
```

PR #478, opened 2018-05-07
Lesson 7

wachumar:
@kuszcjan


students/wachulec_maria/lesson_07_string_datetime/to_swap_the_two_words.py
```diff
@@ -0,0 +1,8 @@
+def swap_two_words():
+    text = input()
+    space_index = text.find(' ')
+    return ' '.join((text[space_index + 1:], text[:space_index]))
```
kuszcjan: will work, but I would personally do it in other way:
```
first, second = input().split()
print("{} {}".format(second, first))
```

PR #482, opened 2018-05-08
Lesson 03

ApusAlfa:
@tomekDiakowTieto


students/marta_herezo/lesson_03/calculator.py
```diff
@@ -0,0 +1,83 @@
...
+
+    elif option == 'h':
+        help()
+
+    elif option == '?':
+        help()
+
+    elif option == 'q':
+        print('GOOD BYE')
+        break
```
tomekDiakowTieto: I think each function should return value :)

PR #486, opened 2018-05-08
Hyska Monika lesson 08

zelazmon:
@adziu 
gotowe


students/hyska_monika/lesson_08_regular_expressions/strong_password_detection.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def password_check(password):
+    length_error = len(password) < 8
+    uppercase_error = re.search(r"[A-Z]", password) is None
+    lowercase_error = re.search(r"[a-z]", password) is None
+    digit_error = re.search(r"\d", password) is None
+    password_ok = not (length_error or digit_error or
+                       uppercase_error or lowercase_error)
+    password_status = {'Password is strong.': password_ok,
```
adziu: You could have started with the dictionary from the beginning:
```
password_status = dict(
    length=len(password) > 8,
    uppercase=re.search(...),
    ...
```
Then you could say, `password_ok = all(password_status)`. You could have messages as values or even exceptions in the other dictionary.

It's also a good idea to keep positive ideas as `True` in boolean variables so `uppercase` instead of `uppercase_error`. This is to avoid confusion when computing several levels of logic like: "not all statuses are different from false". It is better expressed as "any status is false".

students/hyska_monika/lesson_08_regular_expressions/phone_number_validator.py
```diff
@@ -0,0 +1,142 @@
...
+0048 600600600
+"""
+import re
+
+
+def check_phone_number(phone_number):
+    phone_number_regex = re.compile(r'''(
+        (((00|\+)\d{2}\s)|(\((00|\+)\d{2}\)(\s)?))? # country code
+        \d{3}               # first 3 digits
+        ((-|\s)?\d{3}){2}$  # 2 groups: space or - and 3 digits
```
adziu: Should `600-600 600` be valid?

adziu:
:+1: for the tests

PR #489, opened 2018-05-08
Add lesson 8

karinanartowska:


karinanartowska:
@gradopaw please review :)

students/nartowska_karina/lesson_08_regular_expressions/email_validator.py
```diff
@@ -0,0 +1,23 @@
+import re
+
+
+def email_validator(email):
+    r_email = re.compile(r'(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)')
+    if r_email.match(email):
+        return 1
```
gradopaw: Here it is better to return True instead of 1. You should also add explicit return False in a case when the match will be not found. How it is described in One of the ten Zen of Python (https://www.python.org/dev/peps/pep-0020/) Explicit is better than implicit.
```
def email_validator(email):
    r_email = re.compile(r'(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)')
    if r_email.match(email):
        return True
   return False
```

PR #502, opened 2018-05-11
lesson 8

Graheb:
@kamwar Please check. I'm not sure if "strong password" is correct and if I use git correctly.


students/Hebdowska_Grazyna/lesson_08/email_validator.py
```diff
@@ -0,0 +1,19 @@
+import re
+
+
+def email_validator(email):
+
```
kamwar: Unnecessary empty line

students/Hebdowska_Grazyna/lesson_08/email_validator.py
```diff
@@ -0,0 +1,19 @@
...
+    s = r'''((\w{1,})(\.{1})*(\w{1,})*(\w[@]{1})(\w{1,})(\.{1})(\w{2,3}))'''
```
kamwar: {1,} is equivalent to +
{1} might be just omitted

It’s better to use *, +, or ? when you can, simply because they’re shorter and easier to read.

students/Hebdowska_Grazyna/lesson_08/email_validator.py
```diff
@@ -0,0 +1,19 @@
...
+
+    if mo is not None:
+        if mo.group() == email:
+            print(mo.group())
+            return True
+    else:
+        return False
+
+
+test = input("Email:  ")
```
kamwar: Please consider to add py.test for verification.

There is available a perfect feature 'parametrize' for calling the same method many times.

class TestEmailAddressValidator:
    @pytest.mark.parametrize("correct_address", [
        'aaa@aaa.pl',
        'a-a-a@aaa.pl',
        'a12a@12a.com',
    ])
    def test_valid_email_address(self, correct_address):
        assert email_validator(correct_address)

    @pytest.mark.parametrize("incorrect_address", [
        'aaa@a@aa.pl',
        'aaa@@aa.pl',
        'a12a.com',
    ])
    @pytest.mark.xfail
    def test_incorrect_email_address(self, incorrect_address):
        assert email_validator(incorrect_address)

students/Hebdowska_Grazyna/lesson_08/phone_number.py
```diff
@@ -0,0 +1,20 @@
+import re
+
+
+def phone_validator(phone_number):
+
+    s = r'''(\+\d{2})?(\s|-|\.)?(\d{3})(\s|-|\.)?(\d{3})(\s|-|\.)?(\d{3})'''
```
kamwar: delimiter (''') is unnecessary here

students/Hebdowska_Grazyna/lesson_08/strong_pasword.py
```diff
@@ -0,0 +1,30 @@
...
+    re_number3 = re.compile(r'''((\d{1,})+)''')
+    mo1 = re_number1.search(password)
+    mo2 = re_number2.search(password)
+    mo3 = re_number3.search(password)
+    if mo1 is None:
+        return False
+    elif mo2 is None:
+        return False
+    elif mo3 is None:
+        return False
```
kamwar: You might join the if statements:
    if mo1 is None or mo2 is None or mo3 is None:
        return False
    return True

PR #509, opened 2018-05-14
add lesson_09_hyska

zelazmon:
@adziu 
gotowe

### DIFF
--- a/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
+++ b/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
@@ -29,8 +29,8 @@ def set_words_on_file(file_name, word_to_replace):
 
 def prepare_file_name(file_name):
     new_file_name = str(os.path.splitext(file_name)[0]) + \
-                    "_" + str(date.today()) + "_" + str(random.randrange(999)) +\
-                    ".txt"
+                    "_" + str(date.today()) + "_" +\
+                    str(random.randrange(999)) + ".txt"
     return new_file_name
 
 
--- a/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
+++ b/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
@@ -1,0 +1,33 @@
+"""
+Mad Libs program that reads in text files and lets the user add
+their own text anywhere the word ADJECTIVE, NOUN, ADVERB, or
+VERB appears in the text file
+The program would find these occurrences and prompt the user to replace them.
+The results should be printed to the screen and saved to a new text file.
+"""
+import os
+from datetime import date
+import random
+
+
+def set_words_on_file(file_name, word_to_replace):
+    file_text = open(file_name).read()
+    print("My file:\n" + file_text)
+    my_file_without_dots = file_text.replace(".", "")
+    all_words = my_file_without_dots.split(" ")
+    new_file_name = str(os.path.splitext(file_name)[0]) + \
+        "_" + str(date.today()) + "_" + str(random.randrange(999)) + ".txt"
+    with open(new_file_name, "w+") as changed_file:
+        for elem in all_words:
+            if elem in word_to_replace:
+                print("Enter " + elem + ": ")
+                new_word = input(str())
+                file_text = file_text.replace(elem, new_word)
+        changed_file.write(file_text)
+        print("Changed file:\n" + file_text)
+        changed_file.close()
+
+
+my_file_name = "./txt_files/text1.txt"
+base_words = ["ADJECTIVE", "NOUN", "ADVERB", "VERB"]
+set_words_on_file(my_file_name, base_words)
--- a/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
+++ b/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
@@ -28,9 +28,14 @@ def set_words_on_file(file_name, word_to_replace):
 
 
 def prepare_file_name(file_name):
-    new_file_name = str(os.path.splitext(file_name)[0]) + \
+    new_file_name = "{0}_{1}_{2}.txt".format(
+        str(os.path.splitext(file_name)[0]),
+        str(date.today()),str(random.randrange(999)))
+    """
+    new_file_name = str(os.path.splitext(file_name)[0]) +\
                     "_" + str(date.today()) + "_" +\
                     str(random.randrange(999)) + ".txt"
+    """
     return new_file_name
 
 
--- a/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
+++ b/students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
@@ -30,7 +30,7 @@ def set_words_on_file(file_name, word_to_replace):
 def prepare_file_name(file_name):
     new_file_name = "{0}_{1}_{2}.txt".format(
         str(os.path.splitext(file_name)[0]),
-        str(date.today()),str(random.randrange(999)))
+        str(date.today()), str(random.randrange(999)))
     """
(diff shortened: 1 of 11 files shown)


students/hyska_monika/lesson_09_reading_and_writing_files/mcb.pyw
```diff
@@ -0,0 +1,33 @@
...
+        pyperclip.copy(str(list(mcbShelf.keys())))
+    elif sys.argv[1] in mcbShelf:
+        pyperclip.copy(mcbShelf[sys.argv[1]])
+elif len(sys.argv) == 3 and sys.argv[1].lower() == 'delete':
+    if sys.argv[2].lower() == 'all':
+        mcbShelf.clear()
+    elif sys.argv[2] in mcbShelf:
+        del mcbShelf[sys.argv[2]]
+
+mcbShelf.close()
```
adziu: Looks good. It would increase the readability if you named the arguments like `__, command, keyword = sys.argv` instead of pointing at them by index.

students/hyska_monika/lesson_09_reading_and_writing_files/regex_search.py
```diff
@@ -0,0 +1,37 @@
...
+The results should be printed to the screen.
+"""
+
+from os import listdir
+from os.path import isfile, join, os
+import re
+
+
+def files_in_path(path):
+    only_files = [f for f in listdir(path) if isfile(join(path, f))]
```
adziu:
```
import os
...
... os.listdir(path)...
```
is recommended, importing names directly into module is discouraged. The reason for that is that the names will leak into any module that imports this one. Also `os.path.*` is accessible after `os` is imported (some magic).

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+    new_file_name = prepare_file_name(file_name)
+    with open(new_file_name, "w+") as changed_file:
+        for elem in all_words:
+            if elem in word_to_replace:
+                print("Enter " + elem + ": ")
+                new_word = input()
+                file_text = file_text.replace(elem, new_word)
+        changed_file.write(file_text)
+        print("Changed file:\n" + file_text)
+        changed_file.close()
```
adziu: With `with` you don't close the file at the end. It is automatically closed when leaving the context.

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+The program would find these occurrences and prompt the user to replace them.
+The results should be printed to the screen and saved to a new text file.
+"""
+import os
+from datetime import date
+import random
+import re
+
+
+def set_words_on_file(file_name, word_to_replace):
```
adziu: It should be _words_ to replace.

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    new_file_name = str(os.path.splitext(file_name)[0]) +\
+                    "_" + str(date.today()) + "_" +\
+                    str(random.randrange(999)) + ".txt"
+    """
+    return new_file_name
+
+
+def remove_special_characters(file_text):
+    file_without_special_characters = re.sub('[^ a-zA-Z0-9]', '', file_text)
```
adziu: `\w` is a shortcut for alphanumeric, it will also handle Unicode characters like `ą`

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+    file_text = open(file_name).read()
+    print("My file:\n" + file_text)
+    all_words = remove_special_characters(file_text)
+    new_file_name = prepare_file_name(file_name)
+    with open(new_file_name, "w+") as changed_file:
+        for elem in all_words:
+            if elem in word_to_replace:
+                print("Enter " + elem + ": ")
+                new_word = input()
+                file_text = file_text.replace(elem, new_word)
```
adziu: I think I complained about it. You should limit replacement with `file_text.replace(elem, new_word, count=1)` (optional argument `count`), so that it doesn't replace all NOUN occurences.

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+                    "_" + str(date.today()) + "_" +\
+                    str(random.randrange(999)) + ".txt"
+    """
+    return new_file_name
+
+
+def remove_special_characters(file_text):
+    file_without_special_characters = re.sub('[^ a-zA-Z0-9]', '', file_text)
+    # my_file_without_dots = file_text.replace(".", "")
+    all_words = file_without_special_characters.split(" ")
```
adziu: If you do the split at the end, then you are actually producing words... `stripped_words` would be a good name. Also `remove` suggests in-place action `removed` would suggest that you are returning a value.

students/hyska_monika/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,56 @@
...
+            if elem in word_to_replace:
+                print("Enter " + elem + ": ")
+                new_word = input()
+                file_text = file_text.replace(elem, new_word)
+        changed_file.write(file_text)
+        print("Changed file:\n" + file_text)
+        changed_file.close()
+
+
+def prepare_file_name(file_name):
```
adziu: Here is a library for getting files with non-conflicting names: https://docs.python.org/2/library/tempfile.html

I would personally prefer a date *and timestamp* here.

students/hyska_monika/lesson_09_reading_and_writing_files/mcb.pyw
```diff
@@ -0,0 +1,36 @@
...
+py [filename].pyw delete all
+    -- clear all keyword
+"""
+import shelve, pyperclip, sys
+
+
+mcbShelf = shelve.open("mcb")
+keyword = sys.argv
+command = keyword[1].lower()
+sub_command = keyword[2].lower()
```
adziu: Better use unpacking `__, command, subcommand = sys.argv` (`__` stands for ignored). Besides if you try to access `keyword[2]` when `len(keyword) == 2` it should crash with `IndexError`. Have you tested that?

students/hyska_monika/lesson_09_reading_and_writing_files/mcb.pyw
```diff
@@ -0,0 +1,36 @@
...
+
+if len(keyword) == 3 and command == 'save':
+    mcbShelf[keyword[2]] = pyperclip.paste()
```
adziu: You have already unpacked that.

students/hyska_monika/lesson_09_reading_and_writing_files/regex_search.py
```diff
@@ -0,0 +1,37 @@
...
+    files = files_in_path(path)
+    pattern = re.compile(regex)
+    print("Lines contain searched regex:")
+    for file_name in files:
+        file_with_path = os.path.join(path, file_name)
+        file_text = open(file_with_path)
+        for line in file_text:
+            line = line.rstrip()
+            if re.search(pattern, line):
+                print(str(file_name) + ": " + str(line))
```
adziu: Isn't `file_name` a string already?

students/hyska_monika/lesson_09_reading_and_writing_files/regex_search.py
```diff
@@ -0,0 +1,37 @@
...
+        file_text.close()
```
adziu: Better use `with` like before.

PR #512, opened 2018-05-15
Adziu/lesson 03 functions

WinterWerewolf:
@adziu Mam nadzieję, że tym razem jest dobrze. Za bardzo na hurra poszłam...


WinterWerewolf:
Those names are shortcuts I feel comfortable with, e.g. 'nonnegint' means 'NON-NEGative INTeger'. Should I start writing comments explaining those names or try to use more descriptive names? Cause if they are not ok, I will try to be more in line with Python's naming convention.
I am also not sure what do you mean by 'import the code'. I guess all three Collatz exercises could be in one 'Input_Validation' file?
As for the missing exercises - no, please don't merge it yet. I want to finish all the exercises first. Wanted some feedback on the finished ones. Thank you.

adziu:
I got the etymology for `non neg int`, but I found it suboptimal. PEP 8 recommends not shortening words (unless the shortcut is very common) and using underscore if it's not obvious, where to split words. I guess it's not easy to explain on that example. I would just use `n` here, because that's what the exercise description used. Further down the road we have some more meaningful input where we can use names like `adults_count`, `average_age`, `width`, `height` etc.

When it comes to importing the code. If you add an empty `__init__.py` file to the directory, you will be able to say `import collatz` and use `collatz.collatz()`. PEP8 also recommends `lowercase_underscore.py` file names.

WinterWerewolf:
@adziu I finished all exercises. I fixed some things according to your suggestions and tried to be more in line with the variable naming rules. Didn't change old file names though, cause I didn't have the time to check whether git is case sensitive and didn't want to break anything.
Stickler is still having some weird expectations with blank lines and I am not sure if I should be worried about it.

PR #514, opened 2018-05-15
Kuszcjan/lesson 09 reading and writing files

miroslawj:

### DIFF
--- a/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr1.txt
+++ b/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr1.txt
@@ -1,0 +1,5 @@
+kody pocztowe plik nr1
+kod pocztowy Leszno: 64-100, Wroclaw: 54-234, Zielona-Gora: 61-783
+Bydgoszcz 23-75, Opole 56-093
+Katowice 67-632
+
--- a/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr2.txt
+++ b/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr2.txt
@@ -1,0 +1,5 @@
+kody pocztowe plik nr1
+kod pocztowy Poznan: 56-110, Wloclawek 74-224, Jelenia-Gora: 51-383
+Torun 78-452
+Czestochowa 98-231, Krakow 23-564
+
--- a/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/mad_libs.py
+++ b/students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/mad_libs.py
@@ -15,9 +15,4 @@ with open(target_file, 'w') as outfile:
                         new_word = word.replace(replace_word, new_word)
                         break
                 target_line.append(new_word)
-            outfile.write(' '.join(target_line)+'\n')
-
-
-
-
-
+            outfile.write(' '.join(target_line) + '\n')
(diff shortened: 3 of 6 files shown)


students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,52 @@
+file = open('file_to_be_modified.txt', 'r')
```
kuszcjan: OK, but The preffered way to operate on files is using context managers: `with` 
Check out : https://stackoverflow.com/questions/9282967/how-to-open-a-file-using-the-open-with-statement

students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,52 @@
...
+        if word_to_be_replaced is None or noun_position < nearest_word_position:
+            nearest_word_position = noun_position
+            word_to_be_replaced = 'NOUN'
+
+    if word_to_be_replaced is not None:
+        print('Enter a', word_to_be_replaced.lower(), ':')
+        new_word = input()
+        source_text = source_text.replace(word_to_be_replaced, new_word, 1)
+
+    starting_search_position += len(new_word)
```
kuszcjan: The algorithm is little odd. 
there is no need to storing cursor position, we can iterate over every word: `for word in line.split():`
I think it would be easier to debug if modifed text would be stored in separate variable ( instead of changing source).
Multiple `if` statements look like poor old c-style switch statement. It's not bad, but looks ugly.

students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/regex_search.py
```diff
@@ -0,0 +1,27 @@
+import os
+import re
+
+
+def collect_arguments(folder_path, regex_expression):
+    regex_search = re.compile(regex_expression)
+    folder_content = os.listdir(folder_path)
+    list_of_text_files = []
+    for file in folder_content:
+        if file[-4:] == '.txt':
```
kuszcjan: little sketchy approach
Better use:
`if filename.endswith(".txt"):`

students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr1.txt
```diff
@@ -0,0 +1,5 @@
+kody pocztowe plik nr1
```
kuszcjan: I think this file belong to previous lesson.
Please remove it from this PR

students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr2.txt
```diff
@@ -0,0 +1,5 @@
+kody pocztowe plik nr1
```
kuszcjan: I think this file belong to previous lesson.
Please remove it from this PR

students/jakielczyk_miroslaw/lesson_09_reading_and_writing_files/folder_with_text_files/text_file_nr1.txt
miroslawj: folder "folder_with_text_files" has been created on purpose, inside this folder there are two txt files, for these files regex_expression is further used

kuszcjan: ok, got it

PR #515, opened 2018-05-16
lesson_10_hyska_monika

zelazmon:
@adziu 
zrobiona


students/hyska_monika/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -0,0 +1,45 @@
...
+    folders_list = []
+    for folder_name, sub_folders, file_names in os.walk(path):
+        folder_size = 0
+        for file_name in file_names:
+            file_name_with_path = os.path.join(folder_name, file_name)
+            if os.path.getsize(file_name_with_path) > size_in_bytes:
+                files_list.append(os.path.abspath(file_name_with_path))
+            folder_size += os.path.getsize(file_name_with_path)
+        if folder_size > size_in_bytes:
+            folders_list.append(os.path.abspath(folder_name))
```
adziu: The content of the for has grown large enough that you could try to subdivide it into a function. PyCharm has great refactoring tools, you just select a piece of code, choose refactor > extract and it suggests parameters and output. The value from subdividing the code is that you can see which of the scopes variables are needed. For example you are not using `sub_folders` at all (you could rename it to `__`, by the way `_` is for gettext in translation tools by convention).

adziu: I assume you are right that they meant folders having more than 100MB files in them _directly_ and that we don't sum over subfolders.

students/hyska_monika/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+import shutil
+import os
+import re
+
+
+def search_prefix_add_gaps(path, prefix):
+    number = 1
+    for file_name in os.listdir(path):
+        file_with_prefix = prefix.search(file_name)
```
adziu: This will also catch names like:
`spam000.txtwhateverspam001.txt` and `notspam001.txttxt`
If you want a prefix, then it should be `match` and if you don't want extra endings, then you need a dollar sign at the end.

students/hyska_monika/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,35 @@
...
+in the numbering (such as if there is a spam001.txt and spam003.txt but no
+spam002.txt). Have the program rename all the later files to close this gap.
+"""
+import shutil
+import os
+import re
+
+
+def search_prefix_add_gaps(path, prefix):
+    number = 1
```
adziu: Use `for number, file_name in enumerate(os.listdir(path), 1)` for that.

students/hyska_monika/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,35 @@
...
+                        os.path.join(path, new_file_name))
+        number += 1
+
+
+my_prefix = re.compile(r'(spam)(\d\d\d)(.txt)')
+search_prefix_add_gaps('.\\folder\\txt_files', my_prefix)
+
+print('')
+my_prefix = re.compile(r'(ada)(\d\d\d)(.doc)')
+search_prefix_add_gaps('.\\folder\\txt_files', my_prefix)
```
adziu: Great use of the library functions, I didn't know about `zfill`. Try to go further and subdivide parts you can name into functions, it should point you towards simplifying the parts and thinking about what you need in a particular scope. I think it should look like `move_file(path, number)`, but if you need more flexibility pass the regex too `move_file(path, match, number)`.

students/hyska_monika/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,42 @@
...
+Copy these files from whatever location they are in to a new folder.
+"""
+import shutil
+import os
+
+
+# function search file on selected folder for selected extension
+# (possible search for 2 extension in the same time)
+# function return list found files
+def search_files_list(path, ext1, ext2=None):
```
adziu: Why not `*extensions`?

students/hyska_monika/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,42 @@
...
+    files_list = []
+    print('Found files:')
+    for folder_name, sub_folders, file_names in os.walk(path):
+        for file_name in file_names:
+            if ext2 is None:
```
adziu: With `extensions` you could just use `for`.

students/hyska_monika/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,42 @@
...
+    for file in files_list:
+        if os.path.isfile(new_location):
+            continue
+        else:
+            shutil.copy(file, new_location)
+
+
+my_files_list = search_files_list('.\\folder', '.pdf', '.JPG')
+# my_files_list = search_files_list('.\\folder', '.pdf')
+copy_files_to_new_location(my_files_list, 'new_folder')
```
adziu: Nice that you subdivide those, now you could pack that in a `main()` function or a  `move_extensions(source, destination, *extensions)` and wrap that in `if __name__ == '__main__':`

students/hyska_monika/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -1,5 +1,5 @@
 """
-Program  walks through a folder tree and searches for
+Program  walks in to a bar tree and searches for
```
adziu: Hey, that was a joke ;)

students/hyska_monika/lesson_10_organizing_files/selective_copy.py
```diff
@@ -26,6 +26,18 @@ def search_files_list(path, ext1, ext2=None):
     return files_list
 
 
+# function search files on selected folder for selected extensions
+def search_files_list2(path, *extensions):
+    files_list = []
+    print('Found files:')
```
adziu: English: "Files found" is how they write.

students/hyska_monika/lesson_10_organizing_files/selective_copy.py
```diff
@@ -26,6 +26,18 @@ def search_files_list(path, ext1, ext2=None):
...
+    for folder_name, sub_folders, file_names in os.walk(path):
+        for file_name in file_names:
+            if file_name.endswith(extensions):
```
adziu: You need a for-loop here. `endswith` accepts a string.

PR #516, opened 2018-05-16
Lesson 10 solutions

synulewar:


synulewar:
@kuszcjan

students/synowiec_krzysztof/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,36 @@
...
+    pattern = re.compile(r'\d{3}')
+
+    if len(files) > 2:
+        prev_n = None
+        gap_found = False
+        for file in files:
+            curr_n = int(pattern.search(file).group())
+            print("current {}".format(curr_n))
+            if not gap_found:
+                if prev_n is not None and prev_n + 1 != curr_n:
```
kuszcjan: I think it would be better to combine `if` from line 23 and 24 into single multiline `if`

PR #518, opened 2018-05-17
lesson_04_hyska

zelazmon:
@adziu 
lekcja 4


students/hyska_monika/lesson_04_unit_testing/distance_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_string_x2(self):
+        self.assertRaises(ValueError, distance.distance(2, 3, 'aoeu', 2))
+
+    def test_string_y2(self):
+        self.assertRaises(ValueError, distance.distance(2, 3, 2, 'aoeu'))
+
+    # Tests it on corner cases:
+    def test_vaule_zero(self):
+        self.assertEqual(distance.distance(2, 1, 2, 1), 0.0)    # Zero length
```
adziu: In case of floating-point numbers you should use `assertAlmostEqual`.

students/hyska_monika/lesson_04_unit_testing/distance_test.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertEqual(distance.distance(3, -2, -1, 7), 9.84886)
+
+    def test_vaule_2(self):
+        self.assertEqual(distance.distance(-10, 2, 5, 12), 18.02776)
+
+    def test_vaule_3(self):
+        self.assertEqual(distance.distance(1, -5, -10, -12), 13.0384)
+
+    # Test that the order of points does not matter
+    def test_order(self):
```
adziu: In all those places, you can just put the comment in the name. `test_distance_does_not_depend_on_point_order`

students/hyska_monika/lesson_04_unit_testing/fibonacci_test.py
```diff
@@ -0,0 +1,44 @@
...
+    def test_values_6(self):
+        self.assertFalse(fibonacci.fib(6) == 5)
+
+    def test_values_19(self):
+        self.assertEqual(fibonacci.fib(19), 4181)
+
+    def test_many_values(self):
+        l = [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
+        for i in range(2, 10):
+            self.assertTrue(fibonacci.fib(i) == l[i])
```
adziu: If any of those cases, it will stop there. It would be a great place to use `pytest` and its `fixtures`, have a look at `pytest`. It's a bit more magical, but you don't need all the boilerplate.

students/hyska_monika/lesson_04_unit_testing/distance.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def distance(x1, y1, x2, y2):
+    try:
+        vector = math.fabs(math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2))
+        vector = round(vector, 5)
+        print("Length of vector is:", vector)
+        return vector
+    except TypeError:
+        print("It isn't integer!")
```
adziu: You can add `raise` afterwards to propagate the exception further.

students/hyska_monika/lesson_04_unit_testing/distance_pytest.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+from distance import distance
+
+
+def test_invalid_string_arguments_to_raises():
```
adziu: You can also use this to fail with a specific case:
```
@pytest.mark.parametrize("args", [
       ('aoeu', 'aoeu', 'aoeu', 'aoeu'),
       ('aoeu', 2, 3, 4),
       (1, 'aoeu', 3, 4),
       (1, 2, 'aoeu', 4),
       (1, 2, 3, 'aoeu'),
])
def test_invalid_string_arguments_to_raises(args):
   with pytest.raises(ValueError):
       distance(*args)
```

PR #519, opened 2018-05-17
Kamwar/lesson 10 organizing files

jemiekam:
@kamwar

### DIFF
--- a/students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
+++ b/students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
@@ -1,6 +1,6 @@
 import os
 
-path = 'C:\\Program Files'
+path = 'Path\\to\\directory'
 
 for foldername, subfolders, filenames in os.walk(path):
     for file in filenames:
--- a/students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
+++ b/students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
@@ -2,11 +2,12 @@ import os
 
 path = 'Path\\to\\directory'
 
-for foldername, subfolders, filenames in os.walk(path):
+for foldername, _, filenames in os.walk(path):
     for file in filenames:
         path_to_file = os.path.join(foldername, file)
-        if os.path.getsize(path_to_file) > 100000000:
+        file_size = os.path.getsize(path_to_file)
+        if file_size > 100000000:
             print('FILENAME: ' + file + ' SIZE: ' +
-                  str(os.path.getsize(path_to_file)) +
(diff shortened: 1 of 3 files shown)


students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+path = 'Path\\to\\directory'
+
+for foldername, subfolders, filenames in os.walk(path):
```
kamwar: For unused variables there is convention to use '_' but is is not stated in PEP8
for foldername, _, filenames in os.walk(path):

students/jemielity_kamil/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+path = 'Path\\to\\directory'
```
kamwar: Just for you information, path = 'Path/to/directory' will also work for any os.path operations.  Then it's cross platform solution

students/jemielity_kamil/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,31 @@
+import os
+import re
+
+prefix = input('Prefix: ')
+suffix = '.txt'
```
kamwar: These are const values and should be UPPER case

```
prefix = input('Prefix: ')
SUFFIX = '.txt'
PATH= 'C:\\Path\\to\\the\folder'
```

students/jemielity_kamil/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,31 @@
...
+        if mo is not None:
+            files_tab.append(file)
+
+files_tab = sorted(files_tab)
+number_of_files = len(files_tab)
+
+new_names = []
+for i in range(1, number_of_files + 1):
+    if i < 10:
+        new_names.append(prefix + '00' + str(i) + suffix)
```
kamwar: It seems a bit too complicated solution, please try to use format instead
```
{:03d}".format(i)
```

students/jemielity_kamil/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,13 @@
+import os
+import shutil
+
+folder_path = 'C:\\Path\\to\\starting\\folder'
+path_to_final_destination = 'C:\\Path\\to\\final\\destination'
```
kamwar: These are const values, please use UPPER case

PR #520, opened 2018-05-17
Adziu/lesson 06 dicts tuples sets args kwargs

RCanDo:
@adziu


students/arkadiusz_kasprzyk/lesson_06_dicts/guess_the_number.py
```diff
@@ -0,0 +1,32 @@
+import random
+
+
+def guess(N):
+    print("Give your guess (number between 1 " +
+          "and {} or one of '0', '?', 'h', 'help' for help):".format(N))
+    guesset = input("").split(" ")
```
adziu: This is ok, but you could have stored the prompt in a variable and then call `input` on it instead of a separate print.

students/arkadiusz_kasprzyk/lesson_06_dicts/guess_the_number.py
```diff
@@ -0,0 +1,32 @@
...
+    if guesset[0].lower() in {'0', '?', 'h', 'help'}:
+        geusset = set()
+    else:
+        guesset = set(guesset) - set("")
+        geusset = set([int(num) for num in guesset])
```
adziu: Also again you can use split without parameters to skip multiple spaces and newlines handling and `map(int, guesset)` is a natural fit here.

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
```diff
@@ -0,0 +1,91 @@
...
+    --------
+    capitalize_word("spam")         # 'Spam'
+    capitalize_word("Spam")         # 'Spam'
+    capitalize_word("_spam")        # '_spam'
+    capitalize_word(" spam")        # ' spam'
+    capitalize_word("_spam eggs")   # '_spam eggs'
+    """
+    word0 = word[0]
+    char0 = ord(word0)
+    if char0 in range(97, 123):
```
adziu: You assume the encoding here, when it's actually a Unicode character, not an ASCII codepoint. There is a method `upper()` and even `capitalize()`...

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
```diff
@@ -0,0 +1,91 @@
...
+    capitalize("spam and eggs")     # 'Spam And Eggs'
+    capitalize(" spam and eggs")    # ' Spam And Eggs'
+    capitalize("spam and eggs ")    # 'Spam And Eggs '
+    capitalize("?sPam aNd egGs ")   # '?sPam ANd EgGs '
+    capitalize("  spam  .eGgs . ")  # '  Spam  .eGgs . '
+    """
+    text_cap = ""
+    a = 0
+    n = len(text)
+    while a < n:
```
adziu: If you used `split()` and `join()`, it would be easier to read.

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
```diff
@@ -0,0 +1,91 @@
...
+    text: str
+
+    Returns
+    -------
+    text_cap: str
+        Each word of `txt` capitalized.
+
+    Examples
+    --------
+    capitalize("spam and eggs")     # 'Spam And Eggs'
```
adziu: Did you know that you can actually run tests from the documentation. There is even a pytest plugin for that: https://docs.pytest.org/en/latest/doctest.html

students/arkadiusz_kasprzyk/lesson_06_dicts/cubes.py
```diff
@@ -0,0 +1,43 @@
...
+if __name__ == "__main__":
+
+    nums = []
+    while len(nums) < 2:
+        nums += input("Give numbers of elements in sets: ").split(" ")
+
+    nums = nums[:2]
+    nums = [int(n) for n in nums]
+
+    for s in cubes(into_set(nums[0], set()), into_set(nums[1], set())):
```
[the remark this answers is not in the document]
RCanDo: This is a matter of what is *considered* to be more basic data structure, list or dictionary. Finally I think it doesn't matter: I may have method for reading into list so that I can also have a method reading *directly* into dictionary, i.e. methods reading the input and returning directly the type I wish to obtain: list or dictionary (or whatever). I don't feel like lists should be anyhow distinguished and created only to be immediately converted into set (notice that into_set() doesn't internally create any list).
After all, I agree that the solution is not quite clear 
but I simply applied DRY for the purpose of this specific script.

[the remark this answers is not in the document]
RCanDo: Yes, the name is not good... However, my idea was to read input into an already existing set (and by default it is an empty set) - now I see it's not reasonable as proper set methods already exist. Silly me... 
Except this one generalisation the function was just a product of DRY for the purpose of this specific script with no intention to make any further use of it - so it's just a bag for all the things I need at this place. Ugly but fast. The way it takes data results from the form of tests in Snakify.

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
RCanDo: Of course there are proper methods, but using them makes the task trivial! What's the point of such excercises!? I did it this way because it was little bit more interesting! (Not a perfect solution but at least sth to think.)

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
RCanDo: I'm not sure as there is a problem of remembering size (number) of spaces between words. See the last 2 examples (I have just added the last). Does split() return the separators list in case of default sep?

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
RCanDo: I know about but don't know details yet. Sure it's simple but... to many "simple" things to grasp :) Thanks for reminding!

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
adziu: If you want to retain spaces, then `split` won't work with default sep (handling newlines, multiple spaces and tabs), but it will work just as good with space only.

students/arkadiusz_kasprzyk/lesson_06_dicts/uppercase.py
adziu: Just saying, it's not equivanent. I don't need any fixes here.

PR #529, opened 2018-05-19
Adziu/lesson 07 strings

marioosh11:


marioosh11:
@adziu please review

students/mariusz_michalczyk/lesson_07_strings/date_printer.py
```diff
@@ -0,0 +1,7 @@
+from datetime import datetime
+
+
+if __name__ == '__main__':
+    now = datetime.now()
+    print("Today is {0} year {1} month and {2} day. Time is: {3}".
```
adziu: You can use strftime for getting month name also, which would fit "human readable" description a bit more.

students/mariusz_michalczyk/lesson_07_strings/num_words.py
```diff
@@ -0,0 +1,3 @@
+if __name__ == '__main__':
+    input_str = input()
+    print(input_str.count(' ') + 1)
```
adziu: It works on all the test cases and it's an obvious way to use count, so I guess that's what was meant.

But I would argue that you could have multiple spaces between, if it was production code.

students/mariusz_michalczyk/lesson_07_strings/swap_words.py
```diff
@@ -0,0 +1,3 @@
+if __name__ == '__main__':
+    str = input()
```
adziu: `str` is a builtin

students/mariusz_michalczyk/lesson_07_strings/swap_words.py
```diff
@@ -0,0 +1,3 @@
...
+    print((str.split()[-1] + " " + str.split()[-2]))
```
adziu:
```
first, last = words.split() # it's an advantage if it crashes on input that you didn't expect
print(' '.join([last, first])
```
It's very strange that you conted from the end.

students/mariusz_michalczyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,24 @@
+table_data = [['apples', 'oranges', 'cherries', 'banana'],
+              ['Alice', 'Bob', 'Carol', 'David'],
+              ['dogs', 'cats', 'moose', 'goose']]
+
+
+def get_biggest_len(col):
```
adziu: You are reimplementing `max` here.

students/mariusz_michalczyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,24 @@
...
+    bgst_len = 0
+    for item in col:
+        if len(item) > bgst_len:
+            bgst_len = len(item)
+    return bgst_len
+
+
+def print_table(table_data):
+    col_width = len(table_data[0])
```
adziu: It's column count.

students/mariusz_michalczyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,24 @@
...
+    for i in range(col_width):
+        for col in table_data:
```
adziu: It's actually getting a row.

students/mariusz_michalczyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,24 @@
...
+            bgst_len = get_biggest_len(col)
+            print(str(col[i]).rjust(bgst_len), end=' ')
+        print('')
+
+
+if __name__ == '__main__':
+    print_table(table_data)
```
adziu: You need a maximum word for each column and then to use that to print cells in each row:
```
max_length_by_column = [max([len(word) for word in column]) for column in zip(*table_data)]
```

marioosh11:
@adziu fixes provided

students/mariusz_michalczyk/lesson_07_strings/table_printer.py
```diff
@@ -3,22 +3,23 @@
...
-    bgst_len = 0
-    for item in col:
-        if len(item) > bgst_len:
-            bgst_len = len(item)
-    return bgst_len
-
-
 def print_table(table_data):
-    col_width = len(table_data[0])
+    col_width, max_word_len = get_maxs(table_data)
```
adziu: It's still not what they requested. In the example the longest word is "cherries"(8 letters), but looking at the second column, the words are right adjusted to fit the "Alice", "Carol" and "David" (5 letters), not "cherries". For that you would need a maximum for each column, not a global maximum for the whole table. And you don't need the `col_width`, `enumerate` will do.

marioosh11:
ok so lest omit it, I dont want spent much more time on it

PR #532, opened 2018-05-22
All lesson 7

awulw:
@tomekDiakowTieto


students/adam_wulw/lesson_07_manipulating_strings/table_printer.py
```diff
@@ -0,0 +1,23 @@
+
+def printTable(table):
```
tomekDiakowTieto: Underscore, python is not Java :)

PR #533, opened 2018-05-22
Result of 07 lessons

ssaleta:
@gradopaw catching up ;)

### DIFF
--- a/students/saleta_sebastian/lesson_07_string_datetime/date_calculator.py
+++ b/students/saleta_sebastian/lesson_07_string_datetime/date_calculator.py
@@ -1,0 +1,23 @@
+from random import randint
+from datetime import datetime, timedelta
+
+
+def print_shift_date(years, days, hours, minutes):
+    today_date = datetime.now()
+    output_date = datetime(today_date.year + years,
+                           today_date.month,
+                           today_date.day) + timedelta(days=days,
+                                                       hours=hours,
+                                                       minutes=minutes)
+
+    print('SHIFT DATE: {:{dfmt} {tfmt}}'
+          .format(output_date, dfmt='%d-%m-%Y', tfmt='%H:%M:%S'))
+
+
+def main():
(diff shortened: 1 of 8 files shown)


ssaleta:
@gradopaw now it look sfine

students/saleta_sebastian/lesson_07_string_datetime/to_swap_the_two_words.py
```diff
@@ -0,0 +1,17 @@
+def swap_two_words(user_word):
+    white_space_index = user_word.find(' ')
+    print(white_space_index)
```
gradopaw: This print is not needed - I supposing that is for debug reasons

students/saleta_sebastian/lesson_07_string_datetime/to_swap_the_two_words.py
```diff
@@ -0,0 +1,17 @@
...
+    if white_space_index != -1:
```
gradopaw: According to the task description, it should not use loops and if. I know that this is for input validate but why you don't check if someone didn't input three or more words? In such cases, your script is not working. You can consider something like this:
```
    d = user_word.split()
    return ' '.join(d[::-1])
```

This will be also working when you will have more than two words.

students/saleta_sebastian/lesson_07_string_datetime/date_calculator.py
```diff
@@ -0,0 +1,23 @@
...
+          .format(output_date, dfmt='%d-%m-%Y', tfmt='%H:%M:%S'))
+
+
+def main():
+    print_shift_date(randint(0, 100), randint(0, 12),
+                     randint(0, 24), randint(0, 60))
+
+
+if __name__ == '__main__':
+    main()
```
gradopaw: Good point with random data 👍

ssaleta:
@gradopaw Your solution is much better :)

PR #536, opened 2018-05-22
Add Lesson 8 - Regular expressions

rafalskowronski:
@tomekDiakowTieto Finally added Lesson 8 - regular expressions

### DIFF
--- a/students/skowronski_rafal/lesson_08_regular_expressions/regex.py
+++ b/students/skowronski_rafal/lesson_08_regular_expressions/regex.py
@@ -1,0 +1,85 @@
+import re
+import copy
+import enum
+
+
+class PostalCodeKind(enum.Enum):
+    POLISH = 1,
+    AMERICAN = 2
+
+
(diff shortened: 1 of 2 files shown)


students/skowronski_rafal/lesson_08_regular_expressions/regex.py
```diff
@@ -0,0 +1,85 @@
...
+    regex = re.compile('^([{0}])*(.*?)([{0}])*$'.format(chars))
+    match = regex.search(text)
+    if match is None:
+        return copy.copy(text)
+
+    return match.group(2)
+
+
+def is_password_strong(text: str) -> bool:
+    regex = re.compile(r'^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{8,})')
```
tomekDiakowTieto: This regexp is long, we can move it to separate value like:
PASSWORD_REGEXP -> later we can use it once again :)

PR #543, opened 2018-05-27
Add lesson 10

karinanartowska:


students/nartowska_karina/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,14 @@
+import shutil
+import glob
+
+
+def main():
+    ext = input('Enter an extension, e.g .txt: ')
+    new_folder_path = "new/"
```
gradopaw: I think that it will be good to create in the current directory  ```new``` folder.  It will make your script will work without any changes in your scripts 😄  What do you think?

students/nartowska_karina/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,38 @@
...
+
+    tab = sorted(tab)
+
+    for i in range(len(tab)):
+        if i < 9:
+            new.append(prefix + '00' + str(i + 1) + suffix)
+        elif i >= 9 & i < 99:
+            new.append(prefix + '0' + str(i + 1) + suffix)
+        else:
+            new.append(prefix + str(i + 1) + suffix)
```
gradopaw: Lines from 22 to 27 can be just replaced by ```new.append('{}{:03d}{}'.format(prefix, (i + 1), suffix))```

students/nartowska_karina/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,14 @@
...
+
+    for path in glob.glob('./**/*' + ext, recursive=True):
```
gradopaw: This solution is ok. ```os.walk()``` mentioned in this lesson can be used as well.

PR #549, opened 2018-05-29
Tomek diakow/lesson 04

MaciekHalas:
@tomekDiakowTieto


students/halas_maciek/lesson_01/collatz_sequence_2_unittest.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+
+def collatz(x):
+    if x % 2 == 0:
+        res = x // 2
+    else:
+        res = 3 * x + 1
+    return res
```
tomekDiakowTieto: You can use:
return x // 2 if x % 2 == 0 else 3  * x  + 1

But it's only TIP because your solution is propably more clear

PR #553, opened 2018-05-30
Tomek diakow/lesson 05

MaciekHalas:
@tomekDiakowTieto


chessboard.py
```diff
@@ -0,0 +1,13 @@
...
+for x in range(m):
+    chessboard.append([])
+    for y in range(n):
+        if (x + y) % 2 != 0:
+            chessboard[x].append('*')
+        else:
+            chessboard[x].append('.')
+
+for row in chessboard:
+    print(' '.join(row))
```
tomekDiakowTieto: We shoudl rename variables like: m, n, x

coma_code.py
```diff
@@ -0,0 +1,8 @@
+def comma_code(list):
+    output = ''
+    for i in range(len(list) - 1):
+        output += (list[i]) + ', '
+    return output + 'and ' + list[-1]
```
tomekDiakowTieto: We can use format method

age_calc.py
```diff
@@ -0,0 +1,17 @@
+def age_calculate(age):
+    average = "There are no adults in the list."
+    amount_of_children = [i for i in age if i in range(1, 18)]
+    amount_of_adults = [i for i in age if i not in range(18)]
```
tomekDiakowTieto: I think we can use: if i < 18

PR #555, opened 2018-05-30
Add scripts for lessons 08 regex

marioosh11:


marioosh11:
@adziu please review

students/mariusz_michalczyk/lesson_08_regex/email_validator.py
```diff
@@ -0,0 +1,15 @@
+import re
+
+
+def validate_email(email):
+    rgx_email = re.compile(r'[a-zA-Z0-9._%+-]+@[A-Za-z0-9]+(\.[a-zA-Z]{2,4})')
+    matched = rgx_email.search(email)
```
adziu: `validate_email('####valid@email.com###')` will pass, because you used `search` instead of `match` (alternatively you didn't prepend the regex with `^`). The invalid characters at the end will also match with `match`, you need `$` at the end.

students/mariusz_michalczyk/lesson_08_regex/phone_nr_validation.py
```diff
@@ -0,0 +1,15 @@
+import re
+
+
+def validate_phone_nr(nr):
+    regex = re.compile(r'(\+\d{2})*\s*\-*(\d{3})\s*\-*(\d{3})\s*\-*(\d{3})')
```
adziu: `+48 +48  +11 +11   ---   111   ---111      ------111` will match. For 0 or 1 repetitions use `?`, for the dashes you can either do `[ -]` or `( |-)`.

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
```diff
@@ -0,0 +1,16 @@
+import re
+
+
+def strip_mm(str, replacement=''):
```
adziu: `str` is a builtin, don't overwrite

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
```diff
@@ -0,0 +1,16 @@
...
+    if not replacement:
+        rgx_strip = re.compile(r'\s*(\d*\w*)\s*')
+        matched = rgx_strip.search(str)
+        return matched.group(1)
+    else:
+        regex = re.compile(r'^[{0}]*|[{0}]*$'.format(replacement))
```
adziu: Almost, but you need to escape special characters, `re.escape(replacement)`, otherwise `.` will catch everything.

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
```diff
@@ -0,0 +1,16 @@
+import re
+
+
+def strip_mm(str, replacement=''):
+    if not replacement:
+        rgx_strip = re.compile(r'\s*(\d*\w*)\s*')
```
adziu: You can't assume the group won't have spaces: `strip('   aaa   aaa   ') == 'aaa   aaa'`. 

One way is to do the same as you did with replacement, but use `\s` (I'm not sure about newline characters... maybe `[\n\s]`) instead of `[{0}]`.

Another is to actually match everything in the group, but not eagerly, because that would catch the spaces at the end too, i.e. `(.*?)` (the `*?` stands for match many, but as little as possible).

students/mariusz_michalczyk/lesson_08_regex/strong_password_detection.py
```diff
@@ -0,0 +1,15 @@
+import re
+
+
+def check_password(password):
+    rgx_strip = re.compile(r'[0-9]+[a-zA-Z]{8,}|[a-zA-Z]{8,}[0-9]+')
```
adziu: `abcd3aBc` should pass and won't. As they suggested, the easiest way is to use multiple validatiors. Say:
```
len_ok = len(pass) >= 8
digits_ok = re.search(r'\d', pass)
case_ok = ...
conditions = len_ok, digits_ok, case_ok
if all(conditions):
...
```

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
marioosh11: written test prove that it works.

marioosh11:
@adziu fixes delivered, please review

students/mariusz_michalczyk/lesson_08_regex/phone_nr_validation.py
```diff
@@ -1,15 +1,44 @@
...
+        self.assertEqual(
+            validate_phone_nr("+48 999 111 22x"), None)
+
+    def test_mail_missing_digit(self):
+        self.assertEqual(
+            validate_phone_nr("+48 999 111 22"), None)
+
+    def test_mail_additional_space(self):
+        self.assertEqual(validate_phone_nr
+                         ("+48 +48 +11 +11 --- 111 ---111 ------111"), None)
```
adziu: :rofl:

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
```diff
@@ -1,16 +1,49 @@
...
+import unittest
 
 
-def strip_mm(str, replacement=''):
+def strip_mm(text, replacement=''):
     if not replacement:
-        rgx_strip = re.compile(r'\s*(\d*\w*)\s*')
-        matched = rgx_strip.search(str)
+        rgx_strip = re.compile(r'^\s*(\d*\w*)\s*$')
+        matched = rgx_strip.search(text)
```
adziu: Still `match` would be clearer here, because search is meant to give you multiple results.

students/mariusz_michalczyk/lesson_08_regex/regex_strip.py
```diff
@@ -1,16 +1,49 @@
...
+        self.assertEqual(
+            strip_mm(".$leon.$", replacement=".$"), 'leon')
+
+    def test_strip_two_chars_replacement_with_space(self):
+        self.assertEqual(
+            strip_mm("xo takie tam xo", replacement="xo"), ' takie tam ')
+
+    def test_strip_two_chars_replacement_with_space2(self):
+        self.assertEqual(
+            strip_mm("xo takie tam xo ", replacement="xo "), 'takie tam')
```
adziu: In any of the tests you are not checking for different sequences. `xoaaaaox` should also be stripped, but you are always testing `SAMEaaaSAME` and not `AMSEaaaSEMA` for example.

adziu: Looking at the code it should work, but keep that in mind when testing. Try to break things!

PR #566, opened 2018-06-04
Gradopaw/lesson 07 manipulating strings

chylamal:
@gradopaw


students/chylak_malgorzata/lesson_07_manipulating_strings_+_datetime_formatting/to_swap_the_two_words.py
```diff
@@ -0,0 +1,4 @@
+s = input()
+first_word = s[:s.find(' ')]
+second_word = s[s.find(' ') + 1:]
+print(second_word + ' ' + first_word)
```
gradopaw: You can consider as well solution that is working for more than two words:
```
data = input(str()).split(" ")
print((' ').join(data[::-1]))
```

PR #567, opened 2018-06-04
Lesson 10

wachumar:
@kuszcjan

### DIFF
--- a/students/wachulec_maria/lesson_10_organizing_files/deleting_unneeded_files.py
+++ b/students/wachulec_maria/lesson_10_organizing_files/deleting_unneeded_files.py
@@ -1,0 +1,13 @@
+import os
+
+
+def search_large_files(folder_tree):
+    for folders, subfolders, files in os.walk(folder_tree):
+        for i in files:
+            if os.path.getsize(os.path.join(folders, i)) > 100000000:
+                print("Size: {} File: {}".format(os.path.getsize(
+                    os.path.join(folders, i)), os.path.join(folders, i)))
+
+
+location = 'C:\\Users\\wachumar\\Desktop\\książki'
+search_large_files(location)
--- a/students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
+++ b/students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
@@ -41,11 +41,11 @@ def change_name(list_of_all, new_names, base_path):
         print("Renaming...")
         for i in range(len(list_of_all)):
             shutil.move(os.path.join(base_path, list_of_all[i]),
-                        os.path.join(base_path, 'r_'
-                                     + new_names[list_of_all[i]]))
+                        os.path.join(base_path, 'r_' +
+                                     new_names[list_of_all[i]]))
         for i in range(len(list_of_all)):
-            shutil.move(os.path.join(base_path, 'r_'
-                                     + new_names[list_of_all[i]]),
+            shutil.move(os.path.join(base_path, 'r_' +
+                                     new_names[list_of_all[i]]),
                         os.path.join(base_path, new_names[list_of_all[i]]))
     else:
         print("Cancel")
--- a/students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
+++ b/students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
@@ -22,16 +22,15 @@ def search_prefix_in_files_names(prefix):
 def create_new_names(surfix_len, start_surfix, names_list, split_text_list,
                      gap_in):
     new_names, step = {}, 1
-    for i in range(len(names_list)):
+    for i, name in enumerate(names_list):
         if gap_in == i + step:
             step += 1
         amount_null = surfix_len - len(str(start_surfix + i))
-        new_names[names_list[i]] = (split_text_list[i][1] + '0' *
-                                    amount_null + str(i + step) +
-                                    split_text_list[i][3])
+        new_names[name] = (split_text_list[i][1] + '0' *
+                           amount_null + str(i + step) + split_text_list[i][3])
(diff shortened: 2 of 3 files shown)


students/wachulec_maria/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -0,0 +1,13 @@
...
+
+def search_large_files(folder_tree):
+    for folders, subfolders, files in os.walk(folder_tree):
+        for i in files:
+            if os.path.getsize(os.path.join(folders, i)) > 100000000:
+                print("Size: {} File: {}".format(os.path.getsize(
+                    os.path.join(folders, i)), os.path.join(folders, i)))
+
+
+location = 'C:\\Users\\wachumar\\Desktop\\książki'
```
kuszcjan: please avoid hardcoded paths.
prefer os.path.join, to make cross-platform compatible

students/wachulec_maria/lesson_10_organizing_files/deleting_unneeded_files.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+
+def search_large_files(folder_tree):
+    for folders, subfolders, files in os.walk(folder_tree):
+        for i in files:
+            if os.path.getsize(os.path.join(folders, i)) > 100000000:
```
kuszcjan: would be much better to export values like 100000000 into separate var, preferably CONST

students/wachulec_maria/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,27 @@
+import os
+import shutil
+
+
+def solve_conflicts(file_name, new_file, extension):
+    copy_file_name = file_name
+    while copy_file_name in os.listdir(new_file):
+        copy_file_name = copy_file_name[:-len(extension)] + str(1) + extension
```
kuszcjan: why `str(1)` ? why not `'1'` 
Besides, what if even such file exist ?

students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,67 @@
...
+                                 list_of_all_groups, place_of_gap)
+    change_name(names_list, new_names, PATH)
+
+
+PATH = os.path.abspath('./files')
+# RUN to close gap:
+run_filling_in_the_gaps('spam')
+
+# # RUN to add gap:
+# run_filling_in_the_gaps('spam', 3)
```
kuszcjan: kind of strange naming. 
The same function elimates the gap and creates a gap, by providing parameter.

Better approach is to have separate functions.
Goal is to be as readable as possible, not to have fever lines of code or fever functions

students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,67 @@
...
+        if result is None:
+            continue
+        names_list.append(result.group(1))
+        all_groups_list.append(list(result.groups()))
+        list_of_numbers.append(result.group(3))
+    if len(list_of_numbers) == 0:
+        raise AttributeError("There are no file in the path")
+    return len(result.group(3)), int(min(list_of_numbers)), sorted(
+        names_list, reverse=False), all_groups_list
+
```
kuszcjan: kind of overcomplicated return.
When you just look at this function, it is not clear what the return values exactly are.

students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,67 @@
+import os
+import re
+import shutil
+
+
+def search_prefix_in_files_names(prefix):
+    all_groups_list, list_of_numbers, names_list, result = [], [], [], None
```
kuszcjan: when we are creating multiple var in single line, they should be the same type. so here it would be more like:
```
all_groups_list, list_of_numbers, names_list = [], [], []
result = None
```

students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,67 @@
...
+    for file_name in os.listdir(PATH):
+        result = re.search(re.compile('((' + prefix + ')(\d\d\d)(.*?)$)'),
+                           file_name)
+        if result is None:
+            continue
+        names_list.append(result.group(1))
+        all_groups_list.append(list(result.groups()))
+        list_of_numbers.append(result.group(3))
+    if len(list_of_numbers) == 0:
```
kuszcjan: I think it can be simplified to ` if list_of_numbers: `

students/wachulec_maria/lesson_10_organizing_files/filling_in_the_gaps.py
```diff
@@ -0,0 +1,67 @@
...
+        raise AttributeError("There are no file in the path")
+    return len(result.group(3)), int(min(list_of_numbers)), sorted(
+        names_list, reverse=False), all_groups_list
+
+
+def create_new_names(surfix_len, start_surfix, names_list, split_text_list,
+                     gap_in):
+    new_names, step = {}, 1
+    for i in range(len(names_list)):
```
kuszcjan: instead of using ` range(len(` try to do it with `enumerate`

students/wachulec_maria/lesson_10_organizing_files/selective_copy.py
wachumar: I wanted to use str(arg) and asked user about arg but I changed my mind. When is str(1) than it's easier to make this extension.

wachumar:
.

PR #578, opened 2018-06-08
Adziu/lesson 07 string datetime

RCanDo:
@adziu


students/arkadiusz_kasprzyk/lesson_07_strings/date_calculator.py
```diff
@@ -0,0 +1,26 @@
...
+
+def time_from_now(years=0, days=0, hours=0, minutes=0):
+    """
+    Examples
+    --------
+    time_from_now()
+    time_from_now(2, 33, 12, 3)
+    """
+    date0 = dt.datetime.now()
+    date1 = date0.replace(date0.year + years) + \
```
adziu: Pretty solution.

students/arkadiusz_kasprzyk/lesson_07_strings/number_of_words.py
```diff
@@ -0,0 +1,11 @@
+# number_of_words.py
+
+
+def words_number(string):
+    nr = string.count(" ") + 1
```
adziu: That may fail in many ways. If you use `len(string.split())`, then it will handle newlines, tabs, and multiple spaces at least. There may still be dashes, that would fail, but for that we would probably need a regex, not in this lesson.

students/arkadiusz_kasprzyk/lesson_07_strings/swap_words.py
```diff
@@ -0,0 +1,11 @@
+# swap_words.py
+
+
+def swap_words(string):
+    k = string.find(" ")
+    return string[(k + 1):] + " " + string[:k]
```
adziu: A clearer way would be:
```
first, second = string.split(' ')
return ' '.join([second, first])
```

students/arkadiusz_kasprzyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,28 @@
...
+    for k in range(len(table)):
+        lengths = []
+        for s in table[k]:
+            lengths += [len(s)]
+        col_widths += [max(lengths)]
+
+    for r in range(len(lengths)):       # rows
+        for c in range(len(table)):     # columns
+            print(table[c][r].ljust(col_widths[c], fill), end=sep)
+        print()
```
adziu: You are the first one to do that right on the first go.

students/arkadiusz_kasprzyk/lesson_07_strings/table_printer.py
```diff
@@ -0,0 +1,28 @@
...
+    printTable(tableData, sep='...')
+    printTable(tableData, sep='...', fill='.')
+    """
+
+    col_widths = []
+    for k in range(len(table)):
+        lengths = []
+        for s in table[k]:
+            lengths += [len(s)]
+        col_widths += [max(lengths)]
```
adziu: When adding single element use `append()`

students/arkadiusz_kasprzyk/lesson_07_strings/number_of_words.py
RCanDo: That's all true. It was written many weeks ago when I didn't know that .split() is so "intelligent".

students/arkadiusz_kasprzyk/lesson_07_strings/swap_words.py
RCanDo: I cannot disagree :) But it's opportunity to ask which is faster?

adziu: `timeit`

RCanDo: Using `.split()` and `.join()` gives
`486 ns ± 1.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)`
while the previous solution:
`522 ns ± 7.32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)`
Case solved: use methods - they are clearer and faster.
Finally I have made all corrections and now... I'm waiting for the commit :)

PR #593, opened 2018-06-15
Kuszcjan/lesson 12 debugging

mielcszy:
@kuszcjan: ready, please review.

### DIFF
--- a/students/mielczarek_szymon/lesson_12_debugging/coin_toss_debugging.py
+++ b/students/mielczarek_szymon/lesson_12_debugging/coin_toss_debugging.py
@@ -1,0 +1,17 @@
+#! python3
+
+import random
+guess = ''
+while guess not in ('heads', 'tails'):
+    print('Guess the coin toss! Enter heads or tails:')
+    guess = input()
+toss = random.randint(0, 1) # 0 is tails, 1 is heads
+if toss == guess:
+    print('You got it!')
+else:
+    print('Nope! Guess again!')
+    guesss = input()
+    if toss == guess:
+       print('You got it!')
+    else:
+        print('Nope. You are really bad at this game.')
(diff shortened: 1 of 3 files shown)


students/mielczarek_szymon/lesson_12_debugging/coin_toss_debugging.py
```diff
@@ -0,0 +1,24 @@
+#! python3
+import random
+
+COIN_TOSS = {'tails': 0,
```
kuszcjan: IMHO should be singular version: `tail` and `head`.
super minor stuff :)

students/mielczarek_szymon/lesson_12_debugging/combined_validator.py
```diff
@@ -0,0 +1,115 @@
...
+
+    shutil.move(TMPFILE, USERDATA_CSV_FILE)
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Combined Validator program.')
+    parser.add_argument('-v', '--verbosity',
+                        help="Verbosity level. 0 - disabled, 1 - warnings and errors, "
+                             "2 - info. Warnings and errors are enabled by default.",
+                        type=int, choices=(0, 1, 2), default=1)
```
kuszcjan: nice

students/mielczarek_szymon/lesson_12_debugging/combined_validator.py
```diff
@@ -0,0 +1,115 @@
...
+        is_data_valid = True
+        try:
+            combined_validator(email, password, phone_number, postal_code)
+        except ValidatorError as err:
+            logger.error('\n' + str(err))
+            key = input("\nDo you want to try again? Type yes or no: ")
+            if key == "yes":
+                continue
+            is_data_valid = False
+        break
```
kuszcjan: not a fan of this approach. Instead of `while True` and `break` at the end of loop would go with smth like `while is_data_valid:`

Above all try to make code simple and readable as much as possible.

PR #594, opened 2018-06-17
Add solutions for lesson_13

mielcszy:
@kuszcjan: ready, please review.
The math for circle program lazily taken from given solution :)


students/mielczarek_szymon/lesson_13_objects_and_classes/selective_shallow_compare.py
```diff
@@ -0,0 +1,53 @@
...
+                          .format(attr))
+            return False
+    return True
+
+
+def main():
+    p1 = Point()
+    p1.x = 5.0
+    p1.y = 4.0
+    p2 = copy.copy(p1)
```
kuszcjan: this is shortcut, you should define p2 separately  ;p

PR #599, opened 2018-06-20
Kuszcjan/lesson 05

dychomon:
@kuszcjan please check


students/dychowicz_monika/lesson_05/lesson_05.py
```diff
@@ -0,0 +1,99 @@
...
+
+    if len(adults) > 0:
+        avg_adults = sum(adults) / len(adults)
+    else:
+        avg_adults = "No adults"
+    print("The average age of adults is: ", avg_adults)
+    print("Number of children: " + str(len(children)))
+
+
+def main():
```
kuszcjan: what's the purpose of main when `if __name__ == '__main__':` exist ?

students/dychowicz_monika/lesson_05/lesson_05.py
```diff
@@ -0,0 +1,99 @@
...
+    # print(age_calculator(ages))
+    swap_the_columns()
+
+
+if __name__ == '__main__':
+    # greater_than_neighbours()
```
kuszcjan: should not commit code with comments

students/dychowicz_monika/lesson_05/lesson_05.py
```diff
@@ -0,0 +1,99 @@
+def greater_than_neighbours():
```
kuszcjan: algorithm is broken.
does not output valid number for some inputs, e.g `5 4 3 2 1` (should be 0, not 3)

students/dychowicz_monika/lesson_05/lesson_05.py
```diff
@@ -0,0 +1,99 @@
...
+            if new_list[i] > new_list[i + 1]:
+                quanity += 1
+    print("Numbers of elements greater than neighbours: ", quanity)
+
+
+def swap_min_and_max():
+    a = [int(s) for s in input("Give a list of unique integer  numbers seperate by space: ").split()]
+    index_of_min = 0
+    index_of_max = 0
+    for i in range(1, len(a)):
```
kuszcjan: iterating over all to just find max values is not optimal solution.
it's better to use `max` and `min` built-in functions.
could use it with conjunction with `.indexof`

PR #610, opened 2018-06-27
lesson 12

wachumar:
@kuszcjan

### DIFF
--- a/students/wachulec_maria/lesson_12_debugging/validated_user_base.py
+++ b/students/wachulec_maria/lesson_12_debugging/validated_user_base.py
@@ -1,0 +1,124 @@
+import logging
+import re
+import argparse
+import sys
+
+
+def logger_lever_chooser(level):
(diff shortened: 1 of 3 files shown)


students/wachulec_maria/lesson_12_debugging/validated_user_base.py
```diff
@@ -0,0 +1,124 @@
...
+        email_validator(email)
+        logging.info('Correct email')
+        logging.info('Passport validation...')
+        strong_password_detection(password)
+        logging.info('Correct passport')
+        logging.info('Phone number validation...')
+        phone_number_validator(phone_number)
+        logging.info('Correct phone number')
+        update_csv([email, password, phone_number])
+    except Exception:
```
kuszcjan: ALWAYS try to catch specific exceptions. Ones that can be handled in peacefull manner.

PR #618, opened 2018-06-28
Kuszcjan/lesson 06

dychomon:
@kuszcjan resolved 7 tasks from lesson_06


students/dychowicz_monika/lesson_06/accessrights.py
```diff
@@ -0,0 +1,26 @@
+def check_rights():
+    n = int(input("Enter number of files: "))
+    rights = {}
+    for i in range(n):
+        f_line = input("Enter file name and allowed\
```
kuszcjan: bad line breaking. next line should not go so much to the left.

students/dychowicz_monika/lesson_06/accessrights.py
```diff
@@ -0,0 +1,26 @@
...
+    m = int(input("Enter number of operations: "))
+    for i in range(m):
+        action = input(
+            "Enter operation and affected file\
+ (e.g. \"read file.txt\"): ").split()
+        if action[0] == "write":
+            action[0] = "W"
+        elif action[0] == "read":
+            action[0] = "R"
+        elif action[0] == "execute":
```
kuszcjan: could store mapping "X" to "execute" in separete dictionary.
This would eliminate need for ugly elif construct

PR #625, opened 2018-06-29
Kuszcjan/lesson 07

dychomon:


dychomon:
@kuszcjan plz review lesson07

students/dychowicz_monika/lesson_07/TheNumberOfWords
```diff
@@ -0,0 +1,7 @@
+def main():
```
kuszcjan: forgot filename extension .py ?

students/dychowicz_monika/lesson_07/ToSwapTwoWords
```diff
@@ -0,0 +1,7 @@
+def main():
```
kuszcjan: again file extentsion ..

students/dychowicz_monika/lesson_07/TheNumberOfWords
```diff
@@ -0,0 +1,7 @@
...
+    text = input()
+    print(1 + (text.count(' ')))
```
kuszcjan: this won't works.
for example for `input ='one          two '` output should be 2, where it is now 12

PR #628, opened 2018-06-29
lesson_05

jarosnat:
@gradopaw


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_05/swap_min_and_max.py
```diff
@@ -0,0 +1,21 @@
+def func():
+    a = [int(s) for s in input().split()]
+    index_of_min = 0
+    index_of_max = 0
+
+    for i in range(1, len(a)):
+        if a[i] > a[index_of_max]:
+            index_of_max = i
+        if a[i] < a[index_of_min]:
```
gradopaw: In such cases when you have to check multiple conditions but only one could be true better is to use ```elif``` statement. It could stay as it is now :smile

students/jarosz_natalia/lesson_05/age_calculator.py
```diff
@@ -0,0 +1,25 @@
+def average_age(list_of_people):
+    adult_numbers = 0
+    adult_age_summary = 0
+    child_numbers = 0
+
+    for el in list_of_people:
```
gradopaw: instead of cast el to integer twice in line 7 and line 8 better is cast el variable here or just declare in the line 19 just ```list_of_people = [2, 0, 58, 38]```

PR #629, opened 2018-06-29
Result of lesson 09.

ssaleta:


students/saleta_sebastian/lesson_09_reading_and_writing_files/mad_libs.py
```diff
@@ -0,0 +1,21 @@
+def mad_libs(adjective, noun, verb, noun2):
```
gradopaw: According to this task, you should have txt file that contain example text ```The ADJECTIVE panda: ....````. In the next step, you should open the file, read each line and check each word. In a case when a word is equal to one of the ADJECTIVE, NOUN, ADVERB, or VERB you should change this word to this passed by the user.

students/saleta_sebastian/lesson_09_reading_and_writing_files/regex_search.py
```diff
@@ -0,0 +1,25 @@
...
+        if file.endswith(".txt"):
+            with open(file) as reader:
+                line_index = 0
+                for line in reader:
+                    line_index += 1
+                    if re.search(r'{}'.format(sentence), line):
+                        print('file : {}'.format(file))
+                        print('line {}: {}'.format(line_index, line))
+                    else:
+                        print('Cannot find more sentences in files')
```
gradopaw: Is it printed too many times?

students/saleta_sebastian/lesson_09_reading_and_writing_files/mad_libs.py
ssaleta: I came up with solution re.sub(/bword/b, new_word) check this please in my new commit. If you want I can do this in this way that you write above.

students/saleta_sebastian/lesson_09_reading_and_writing_files/regex_search.py
ssaleta: I think it's ok and output is clearly for user. For example: 
python3 regex_search.py 
NOUN
file : story2.txt
line 1: The ADJECTIVE panda walked to the NOUN and then VERB. A nearby NOUN was unaffected by these events.

file : story.txt
line 1: The ADJECTIVE panda walked to the NOUN and then VERB. A nearby NOUN was unaffected by these events.

ssaleta:
@gradopaw 👍 ❓

PR #632, opened 2018-06-30
Gradopaw/lesson 09

akpoik:
@gradopaw


students/kulon_aleksander/lesson_09/mcb.pyw
```diff
@@ -0,0 +1,61 @@
...
+
+
+def main():
+    mcb_shelf = shelve.open('mcb')
+
+    parser = argparse.ArgumentParser(
+        description='Program to keep track of multiple pieces of text')
+
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('-save',
```
gradopaw: long argument name should be with two dash - not needed to be changed 😄

PR #636, opened 2018-07-02
lesson 10

serekwoj:
@gradopaw lesson 10


students/serek_wojciech/lesson_10_organizing_files/filling_gaps.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
gradopaw: In general, everything is ok, apart from one thing after running this script as a result in test_tree_gaps folder I have not all files numbered properly. Is it working on your side? It looks that it is strictly connected with shutil.move() - you want to move the file to the destination that already exists and shutil.move() cannot deal with this.

PR #637, opened 2018-07-02
new_lesson_12_monika_hyska

zelazmon:
@adziu


students/hyska_monika/lesson_12_Debugging/combined_validator.py
```diff
@@ -0,0 +1,35 @@
...
+    email = input('Put email:')
+    while validators.check_email(email) is False:
+        email = input('Put email:')
+    # ... set password ...
+    password = input('Put password:')
+    while validators.password_check(password) is False:
+        password = input('Put password:')
+    # ... set postcode ...
+    post_code = input('Put postcode: ')
+    while validators.check_postal_code(post_code) is False:
```
adziu: `condition is False` => `not condition`, also is False checks for identity, so an empty list is not False. To get until-loop in Python and not repeat condition, which is dangerous, use `while True: if condition break`

students/hyska_monika/lesson_12_Debugging/csv_functions.py
```diff
@@ -0,0 +1,40 @@
...
+            logging.info('header - columns name: (%s) are saved' % header)
+    with open(filename) as inf:
+        reader = csv.reader(inf.readlines())
+    with open(filename, 'w', newline='') as outf:
+        writer = csv.writer(outf)
+        repalced = 0
+        for line in reader:
+            logging.info('Line in file: %s' % line)
+            if not line:
+                logging.error('IndexError: list index out of range. '
```
adziu: It might be confusing to use exception names, there is `logging.exception` to be used in except section for that. It will save traceback also.

students/hyska_monika/lesson_12_Debugging/validators.py
```diff
@@ -0,0 +1,73 @@
...
+    lowercase_error = re.search(r"[a-z]", password) is None
+    digit_error = re.search(r"\d", password) is None
+    password_ok = not (length_error or digit_error or
+                       uppercase_error or lowercase_error)
+    password_status = {'Password is strong.': password_ok,
+                       'length to short': length_error,
+                       'missing digits': digit_error,
+                       'missing uppercase': uppercase_error,
+                       'missing lowercase': lowercase_error}
+    result = [k for k, v in password_status.items() if v is True]
```
adziu: `is True` is a bad idea, it might bite you.

PR #642, opened 2018-07-02
tasks for lesson_13_objects_and_classes

miroslawj:


students/jakielczyk_miroslaw/lesson_13_objects_and_classes/object_inspector_1.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, name, surname, age, weight):
+        self.name = name
+        self.surname = surname
+        self.age = age
+        self.weight = weight
+
+
+def inspect_object(checked_object):
+    obj_dictionary = checked_object.__dict__
```
kuszcjan: It's worth mentioning that not all object have `__dict__` attribute (e.g. some external libs etc.)
It may not always work, where `getattr` works differently under the hood and can access all attributes

PR #645, opened 2018-07-04
lesson_10

akpoik:
@gradopaw

### DIFF
--- a/students/kulon_aleksander/lesson_10/filling_in_the_gaps.py
+++ b/students/kulon_aleksander/lesson_10/filling_in_the_gaps.py
@@ -2,27 +2,32 @@ import re
 import os
 import shutil
 
-pattern = 'spam00'
+pattern = 'spam'
 path = '../lesson_10/testfolder/'
 
 
 def main():
     sorted_numbers = []
-    prefix_pattern = re.compile('(?<={})\w+'.format(pattern))
+    prefix_pattern = re.compile(r"{0}\d{{3}}".format(pattern))
 
     for folder_name, sub_folder, file_names in os.walk(path):
         for file_name in file_names:
             result = re.search(prefix_pattern, file_name)
             if result is not None:
-                sorted_numbers.append(int(result.group(0)))
+                sorted_numbers.append(result.group(0).replace(pattern, ''))
 
     sorted_numbers = sorted(sorted_numbers)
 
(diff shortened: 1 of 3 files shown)


akpoik:
@gradopaw

students/kulon_aleksander/lesson_10/selective_copy.py
```diff
@@ -0,0 +1,22 @@
...
+def main():
+    if not os.path.exists(path + new_folder):
+        os.makedirs(path + new_folder)
+
+    for root, directory, files in os.walk(path):
+        for file in files:
+            if fnmatch.fnmatch(file, "*.jpg"):
+                file_path = os.path.join(root, file)
+                print(file_path)
+                shutil.copy(file_path, path + new_folder)
```
gradopaw: Is it working correctly? You shouldn't concern files from ```new_folder``` directory as a potentials files to copy. Because of that ```SameFileError``` is raised :smile

students/kulon_aleksander/lesson_10/filling_in_the_gaps.py
```diff
@@ -0,0 +1,29 @@
+import re
+import os
+import shutil
+
+pattern = 'spam00'
```
gradopaw: Because of that program will be not working correctly for files such as spam010.txt or spam100.txt. Shouldn't you consider such files as well? Maybe try to use better regex 😄

akpoik:
@gradopaw

students/kulon_aleksander/lesson_10/filling_in_the_gaps.py
```diff
@@ -19,11 +19,9 @@ def main():
     sorted_numbers = sorted(sorted_numbers)
 
     for i in range(int(sorted_numbers[0]), int(sorted_numbers[-1]) - 1):
-        current = str(sorted_numbers[i-1]).zfill(3)
-        next = str(int(sorted_numbers[i-1]) + 1).zfill(3)
-        later = str(int(sorted_numbers[i-1]) + 2).zfill(3)
+        next = str(int(sorted_numbers[i - 1]) + 1).zfill(3)
```
gradopaw: Try to avoid built-in functions name. ```next``` is one of them https://docs.python.org/3.6/library/functions.html
Just remember about this next time 😄

PR #649, opened 2018-07-05
Lesson 08

Julitt:
@tomekDiakowTieto


students/bedkowska_julita/lesson_08_regex/phone_number_validator.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+
+def check_phone_number(phone):
+    if re.compile(r'^\d{3}-\d{3}-\d{3}$').match(phone):
+        return True
+    return False
```
tomekDiakowTieto: Maybe try in one line ?
return bool(re.compilet(...)

PR #652, opened 2018-07-08
Gradopaw/lesson 11

akpoik:
@gradopaw


students/kulon_aleksander/lesson_11/filling_in_the_gaps.py
```diff
@@ -0,0 +1,47 @@
...
+
+def main():
+    sorted_numbers = []
+
+    parser = argparse.ArgumentParser(description='Filling gaps.')
+
+    parser.add_argument(
+        '--pattern', action="store",
+        dest='pattern',
+        default=pattern,
```
gradopaw: ```default``` flag shouldn't be used with ```required``` flag set to true. It hasn't sense😄 Have a look at this short descriptions: https://docs.python.org/2/library/argparse.html#argparse.ArgumentParser.add_argument

students/kulon_aleksander/lesson_11/filling_in_the_gaps.py
```diff
@@ -0,0 +1,47 @@
...
+        required='True')
```
gradopaw: For ```required``` flag bool is expected not string

students/kulon_aleksander/lesson_11/selective_copy.py
```diff
@@ -0,0 +1,37 @@
...
+new_folder = 'new_folder'
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Selective copy.')
+
+    parser.add_argument(
+        '-d', action="store",
+        dest='new_folder',
+        required='True')
```
gradopaw: bool instead of string

akpoik:
@gradopaw

PR #655, opened 2018-07-09
Gradopaw/lesson 08

jarosnat:
@gradopaw


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_08/regex_version_of strip.py
```diff
@@ -0,0 +1,14 @@
+import re
+
+
+def regex_strip(string, chars_to_remove):
```
gradopaw: You should set the default value for ```chars_to_remove``` because Python strip function by default removes leading and trailing spaces from a string, so it should look like:

```def regex_strip(string, chars_to_remove=' '):```

File: students/jarosz_natalia/lesson_08/regex_version_of strip.py [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+    strip_regex = re.compile('[%s]' % chars_to_remove)
```
gradopaw: Your regex should be a bit rebuilt. Lets consider something like this:
```re.compile("^([" + chars_to_remove+ "]*) | ([" + chars_to_remove+ "]*)$")```

With this small change behaviour of your regex_strip function will be the same as strip() function

**Comment by jarosnat** [comment]
@gradopaw

File: students/jarosz_natalia/lesson_08/regex_version_of_strip.py [hunk 3]
```diff
@@ -0,0 +1,14 @@
+import re
+
+
+def regex_strip(string, chars_to_remove=' '):
+    strip_regex = re.compile("^([" + chars_to_remove + "]*)")
```
gradopaw: Almost ok. You handle here the only case when ```chars_to_remove``` is at the beginning of the string. Now you should do a similar thing for the case when ```chars_to_remove``` is at the end.
Your line 5 should look like:
```strip_regex = re.compile("^([" + chars_to_remove+ "]*) | ([" + chars_to_remove+ "]*)$")```

jarosnat:
@gradopaw

PR #660, opened 2018-07-11
Hyska_Monika_lesson_14

zelazmon:
@adziu


students/hyska_monika/lesson_14_decorators/authorization.py
```diff
@@ -0,0 +1,14 @@
...
+    '''
+    let's pretend this function calls some authorization services
+    returns True if access is granted and False if not
+    access is granted every 3rd time
+    '''
+    global access_count
+    access_count += 1
+    if access_count >= 3:
+        access_count = 0
+    return access_count == 0
```
adziu: You took the instructions very literally. I didn't expect that level of conformance. 

It would be much more readable as:
```
access_count = 0
...
def ...
    access_count += 1
    return access_count in (2, 3)
```

students/hyska_monika/lesson_14_decorators/exercise1.py
```diff
@@ -0,0 +1,21 @@
+"""
+Write @sort decorator that when applied to a function
+that returns a list, sorts this list,
+"""
+
+
+def sort(func):
+    def func_wrapper():
```
adziu: Also take a look at `functools.wraps`: https://stackoverflow.com/a/309000/69769

students/hyska_monika/lesson_14_decorators/exercise1.py
```diff
@@ -0,0 +1,21 @@
...
+        my_list = func()
```
adziu: `result_list` maybe

students/hyska_monika/lesson_14_decorators/exercise1.py
```diff
@@ -0,0 +1,21 @@
...
+        my_list.sort()
```
adziu: The other way that doesn't modify the list in place is `return sorted(my_list)`. Depending on the situation you should use one or the other.

students/hyska_monika/lesson_14_decorators/exercise3.py
```diff
@@ -0,0 +1,22 @@
+"""
+Write @logged decorator using logger_wrapper
+from lesson 6. Apply it to several functions
+to demonstrate that it works
+"""
```
adziu: You forgot about that file probably, below you have the solution to sort i.e. exercise 2.

PR #662, opened 2018-07-12
Added lesson 11 solutions

swietmon:
@adziu Lesson 11 ready for review.

### DIFF
--- a/students/swietczak_monika/lesson_11_argparse_and_math/filling_the_gaps_argparse.py
+++ b/students/swietczak_monika/lesson_11_argparse_and_math/filling_the_gaps_argparse.py
@@ -1,0 +1,42 @@
+import os
+import re
+import argparse
+
+
+def check_args(args=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--prefix",
+                        help="prefix",
+                        default="spam")
+    parser.add_argument("-d", "--directory",
+                        help="directory",
+                        default=r"C:\Python36\gaps")
+    results = parser.parse_args(args)
+    return results.prefix, results.directory
+
+
+def fill_the_gaps(folder):
+    for foldername, subfolders, filenames in os.walk(folder):
+        filenames = sorted(filenames)
+        index = 1
+        new_filename = ""
+        for filename in filenames:
+            if pattern.match(filename):
+                if index == 1:
+                    new_filename = filename
+                print(filename, new_filename, index)
+                if filename != new_filename:
+                    print(
+                        "Lack of filename: " + new_filename + "\nFile " +
+                        filename + " will be renamed.")
+                    filename_path = os.path.join(foldername, filename)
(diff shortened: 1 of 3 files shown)


students/swietczak_monika/lesson_11_argparse_and_math/filling_the_gaps_argparse.py
```diff
@@ -0,0 +1,42 @@
...
+def check_args(args=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--prefix",
+                        help="prefix",
+                        default="spam")
+    parser.add_argument("-d", "--directory",
+                        help="directory",
+                        default=r"C:\Python36\gaps")
+    results = parser.parse_args(args)
+    return results.prefix, results.directory
```
adziu: In real life it would be better if you left that unpacked i.e. `return results` This way you make it easier to extend that to use more parameters: in case other function would use that call you don't have to change that (accessing `results.prefix` will work regardless of the new parameters added). 

Also you avoid possible mistakes from swapping the order. It is general rule that it's better to use keyword parameters and also limit the positional parameters to maximum of three.

students/swietczak_monika/lesson_11_argparse_and_math/filling_the_gaps_argparse.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def fill_the_gaps(folder):
+    for foldername, subfolders, filenames in os.walk(folder):
+        filenames = sorted(filenames)
+        index = 1
+        new_filename = ""
+        for filename in filenames:
+            if pattern.match(filename):
+                if index == 1:
```
adziu: If you fix the lesson 10, you don't have to fix it here.

students/swietczak_monika/lesson_11_argparse_and_math/filling_the_gaps_argparse.py
```diff
@@ -0,0 +1,42 @@
...
+                    filename_path = os.path.join(foldername, filename)
+                    new_filename_path = os.path.join(foldername, new_filename)
+                    os.rename(filename_path, new_filename_path)
+                new_filename = filename.replace(str(index), str(index + 1))
+                index += 1
+
+
+if __name__ == '__main__':
+    prefix, directory = check_args()
+    pattern = re.compile(prefix + r'.*(\d{3})')
```
adziu: It would be better to put it inside the function. It could also be a parameter, but you would have to deal with the assumption about its content. You assume you have a `r'\d{3}'` inside the regex. 

If you didn't use any global values in the beginning it would be easier to integrate here. Don't worry too much about performance in such cases. It's called premature optimalization. You can easily compile the regex each time you call the function in this exercise. Also in many bigger programs that wouldn't be an issue. If it becomes one, THEN you fix it.

PR #671, opened 2018-07-16
lesson_10

jarosnat:
@gradopaw

### DIFF
--- a/students/jarosz_natalia/lesson_10/filling_in_the_gaps.py
+++ b/students/jarosz_natalia/lesson_10/filling_in_the_gaps.py
@@ -1,0 +1,36 @@
+import re
+import os
+import shutil
+
+pattern = 'spam'
+path = '../lesson_10/test/'
+
+
+def main():
+    sorted_numbers = []
+    prefix_pattern = re.compile(r"{0}\d{{3}}".format(pattern))
+
+    for foldername, subfolder, filenames in os.walk(path):
+        for filename in filenames:
+            result = re.search(prefix_pattern, filename)
+            if result is not None:
+                sorted_numbers.append(result.group(0).replace(pattern, ''))
+
+    sorted_numbers = sorted(sorted_numbers)
+
+    for i in range(int(sorted_numbers[0]), int(sorted_numbers[-1]) - 1):
+        next = str(i + 1).zfill(3)
+        later = str(i + 2).zfill(3)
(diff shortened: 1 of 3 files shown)


jarosnat:
@gradopaw

students/jarosz_natalia/lesson_10/filling_in_the_gaps.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for foldername, subfolder, filenames in os.walk(path):
+        for filename in filenames:
+            result = re.search(prefix_pattern, filename)
+            if result is not None:
+                sorted_numbers.append(result.group(0).replace(pattern, ''))
+
+    sorted_numbers = sorted(sorted_numbers)
+
+    for i in range(int(sorted_numbers[0]), int(sorted_numbers[-1]) - 1):
```
gradopaw: It is working but for easiest cases when we have two files spam001.txt and spam003.txt. What is the case when in test folder we will have spam001.txt, spam003.txt and spam005.txt?

jarosnat:
@gradopaw

students/jarosz_natalia/lesson_10/filling_in_the_gaps.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for folder_name, sub_folder, file_names in os.walk(PATH):
+        for file_name in file_names:
+            result = re.search(prefix_pattern, file_name)
+            if result is not None:
+                sorted_numbers.append(result.group(0).replace(PATTERN, ''))
+
+    sorted_numbers = sorted(sorted_numbers)
+
+    for i in range(int(sorted_numbers[0]), int(sorted_numbers[-1]) - 1):
```
gradopaw: This is still not working :) Variables: ```later_number``` and ```next_file``` are referenced before assignment so they causing ```UnboundLocalError```. Please consider something like this (instead of lines 21 - 36)

```
    for i, value in enumerate(sorted_numbers):
        if str(i + 1).zfill(3) == value:
            continue

        next_one = str(i + 1).zfill(3)
        current = PATH + PATTERN + next_one + '.txt'
        nex_one = PATH + PATTERN + value + '.txt'
        if not os.path.isfile(current):
            shutil.move(nex_one, current)
```

jarosnat:
@gradopaw

PR #672, opened 2018-07-17
Results of lesson 13.

ssaleta:


ssaleta:
@gradopaw I'm already in the yard and see the duck

PR #677, opened 2018-07-18
Added lesson 12 solutions

swietmon:


swietmon:
@adziu lesson12 ready for review

students/swietczak_monika/lesson_12_debugging/regex_validators.py
```diff
@@ -0,0 +1,34 @@
...
+def postal_code_validator(some_postal_code):
+    postal_code_regex = re.compile(r"^\d\d-\d\d\d$")
+    if postal_code_regex.match(some_postal_code) is None:
+        return False
+    else:
+        return True
+
+
+def pass_checker(some_password):
+    if re.match(r'^(?=.*[A-Z])(?=.*[a-z])(?=.*[0-9]).{8,}$',
```
adziu: That's an interesting solution! In all the other reviews multiple regular expressions were used. :+1:

students/swietczak_monika/lesson_12_debugging/verbose_output.py
```diff
@@ -0,0 +1,57 @@
...
+        logger.setLevel(logging.WARNING)
+    else:
+        logger.disabled = True
+
+
+def regex_search(reg, folder):
+    regex_text = re.compile(reg)
+    logging.info("Regex = {}".format(regex_text))
+    files_list = os.listdir(folder)
+    logging.info("Files: {}".format(files_list))
```
adziu: It's good to use logger that knows where you are. That is `logger = logging.getLogger(__name__)` at the beginning of the file and then `logger.info(blah)`.

students/swietczak_monika/lesson_12_debugging/regex_validators.py
```diff
@@ -0,0 +1,34 @@
+import re
+
+
+def email_validator(some_email):
+    email_regex = re.compile(r"[^@]+@[^@]+\.[^@]+")
```
adziu: It's more complicated than that. Not all the symbols are allowed. In real life you should search Google for "regex library" and get a comprehensive checker or use some builtins. In less real life situations you should take care that you accept a reasonable subset with `_` `.` etc., but not too much. 

The input you accepted could cause some trouble when. For example didn't put `$` at the end, so the potential attacker could put anything in the content of the mail message you are sending. You also accept spaces/inequality characters, semicolons, all rather dangerous in this context. `[\w\d_.-]+` `[a-z0-9_.-]+` would be much better than `[^@]+` here. In some places the `[^` construct is the best way to go and good that you know it.

PR #689, opened 2018-07-21
Add lesson 13

karinanartowska:


karinanartowska:
@gradopaw please review :)

students/nartowska_karina/lesson_13_objects_and_classes/object_inspector_1.py
```diff
@@ -0,0 +1,20 @@
+class Animal:
+    def __init__(self, type, name, age):
```
gradopaw: ```type``` is built-in function. Try to avoid such variables names 😄

students/nartowska_karina/lesson_13_objects_and_classes/object_inspector_2.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, type, name, age):
+        self.type = type
+        self.name = name
+        self.age = age
+
+
+def dictionary(obj):
+    new_dict = obj.__dict__
+    for j in new_dict.keys():
+        if isinstance(new_dict[j], str):
```
gradopaw: These three ```if/elif``` statement can be combined to the one ```if```:
```
        if isinstance(new_dict[j], (str, int, float)):
            print("key=", j, "value=", new_dict[j])
```

students/nartowska_karina/lesson_13_objects_and_classes/selective_shallow_compare.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, type, name, age):
+        self.type = type
+        self.name = name
+        self.age = age
+
+
+def compare_objects(obj_1, obj_2, atr_names):
+    atr_obj_1 = []
+    atr_obj_2 = []
+    for j in atr_names:
```
gradopaw: instead of comparing two lists it could be done comparing each attribute elements so the program will exit when first attribute will be different without iterating over all attributes and creating a list of them. It can stay as it is now 😄

karinanartowska:
@gradopaw please review :)

PR #700, opened 2018-07-22
Lesson 10

GrzegorzBajorski:
Signed-off-by: Grzegorz Bajorski <Grzegorz.Bajorski@tieto.com>

### DIFF
--- a/students/grzegorz_bajorski/lesson_10_organizing_files/deleting_inneeded_files.py
+++ b/students/grzegorz_bajorski/lesson_10_organizing_files/deleting_inneeded_files.py
@@ -1,0 +1,8 @@
+import os
+
+
+max_size = 1024 * 1024 * 100
+
+for filename in os.listdir():
+    if os.path.getsize(filename) > max_size:
+        print("File ", filename , "size", os.path.getsize(filename))
--- a/students/grzegorz_bajorski/lesson_10_organizing_files/deleting_inneeded_files.py
+++ b/students/grzegorz_bajorski/lesson_10_organizing_files/deleting_inneeded_files.py
@@ -5,4 +5,4 @@ max_size = 1024 * 1024 * 100
 
 for filename in os.listdir():
     if os.path.getsize(filename) > max_size:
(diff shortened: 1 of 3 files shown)


students/grzegorz_bajorski/lesson_10_organizing_files/deleting_inneeded_files.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+max_size = 1024 * 1024 * 100
+
+for filename in os.listdir():
```
gradopaw: ```os.listdir()``` should be used with path argument. In a case when you want to list the current directory it should look:
```os.listdir('.')```

File: students/grzegorz_bajorski/lesson_10_organizing_files/filling_in_the_gaps.py [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+        files_list.append(filename)
+
+files_list.sort()
+
+prev_num = 0
+for files in files_list:
+    current_num = int(files.split("spam")[1].split(".")[0])
+    expected_num = prev_num + 1
+    if current_num != expected_num:
+        new_name = "spam" + str(expected_num) + ".txt"
```
gradopaw: To keep proper file name conventions (spam00X) you should pad string on the left with zeros to fill width using ```zfill(3)```

students/grzegorz_bajorski/lesson_10_organizing_files/selective_copy.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+dest_dir = "copy_folder/"
+os.mkdir(dest_dir)
+
+for filename in os.listdir():
+    if filename.endswith('.pdf'):
+        print(filename)
+        shutil.copy(filename, dest_dir)
+    if filename.endswith('.txt'):
```
gradopaw: instead of the second if better is used to use elif statement, so in the case when first if will be True second one will be not checked. You can also combine this two if statement into one.

PR #710, opened 2018-07-23
lesson 13

wachumar:
@kuszcjan


students/wachulec_maria/lesson_13_objects_and_classes/object_inspector_1.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def give_names_and_values_of_object(obj, attributes):
+    result = {}
+    for name in attributes:
+        result[name] = obj.__getattribute__(name)
+    return result
+
+
+st = Something()
```
kuszcjan: it's better to pass attrs in constructor, not rewrite them later

kuszcjan:
Contratz

PR #719, opened 2018-07-26
lesson_08

MaciekHalas:
@tomekDiakowTieto

### DIFF
--- a/students/halas_maciek/email.py
+++ b/students/halas_maciek/email.py
@@ -1,0 +1,17 @@
+import re
+
+
+def check_email(email_address):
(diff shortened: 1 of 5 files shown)


students/halas_maciek/email.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+
+def check_email(email_address):
+    mail_regex = re.compile(r'^[a-zA-z0-9+._-]+[@][a-zA-Z0-9-]+[.][a-z]+$')
+    if mail_regex.match(email_address):
```
tomekDiakowTieto: Maybe return bool(mail_regex.match(...)

and maybe we should use this regex as variable:
mail_pattern = r'^[a-zA-z0-9+._-]+[@][a-zA-Z0-9-]+[.][a-z]+$'

